PR #4, opened 2015-01-12
Logic for default empty constructor is fixed

Litee:
Logic worked differently between Release and Debug mode - now fixed

### DIFF
--- a/src/ApiApprover/Properties/AssemblyInfo.cs
+++ b/src/ApiApprover/Properties/AssemblyInfo.cs
@@ -1,5 +1,4 @@
 ﻿using System.Reflection;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 // General Information about an assembly is controlled through the following 
--- a/src/ApiApprover/PublicApiGenerator.cs
+++ b/src/ApiApprover/PublicApiGenerator.cs
@@ -8,9 +8,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.CSharp;
 using Mono.Cecil;
-using Mono.Cecil.Cil;
 using Mono.Cecil.Rocks;
-using Mono.Collections.Generic;
 using TypeAttributes = System.Reflection.TypeAttributes;
 
 // ReSharper disable CheckNamespace
@@ -432,7 +430,7 @@ namespace ApiApprover
 
         private static void AddCtorToTypeDeclaration(CodeTypeDeclaration typeDeclaration, MethodDefinition member)
         {
-            if (member.IsAssembly || member.IsPrivate || IsEmptyDefaultConstructor(member))
+            if (member.IsAssembly || member.IsPrivate)
                 return;
 
             var method = new CodeConstructor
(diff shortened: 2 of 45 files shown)


JakeGinnivan:
Is there a way to detect the opcodes injected by nCrunch and remove. I guess if not we can just ignore the test with nCrunch but would prefer to keep nCrunch support if possible

Litee:
Technically possible, but requires some code - here is how IL for empty method looks like:

```
+		[0]	{IL_0000: nop}	Mono.Cecil.Cil.Instruction
+		[1]	{IL_0001: ldc.i4.m1}	Mono.Cecil.Cil.Instruction
+		[2]	{IL_0002: stloc V_0}	Mono.Cecil.Cil.Instruction
+		[3]	{IL_0006: ldc.i4.0}	Mono.Cecil.Cil.Instruction
+		[4]	{IL_0007: stloc V_1}	Mono.Cecil.Cil.Instruction
+		[5]	{IL_000b: call System.Void nCrunch.TestRuntime.TestCoverageEventListener::NCrunchEnsureStackSpaceExists()}	Mono.Cecil.Cil.Instruction
+		[6]	{IL_0010: ldc.i4 1}	Mono.Cecil.Cil.Instruction
+		[7]	{IL_0015: ldc.i4 598}	Mono.Cecil.Cil.Instruction
+		[8]	{IL_001a: ldloca V_0}	Mono.Cecil.Cil.Instruction
+		[9]	{IL_001e: ldloca V_1}	Mono.Cecil.Cil.Instruction
+		[10]	{IL_0022: call System.Void nCrunch.TestRuntime.TestCoverageEventListener::NCrunchMarkTestLineStart(System.Int32,System.Int32,System.Int32&,System.UInt32&)}	Mono.Cecil.Cil.Instruction
+		[11]	{IL_0027: ldarg.0}	Mono.Cecil.Cil.Instruction
+		[12]	{IL_0028: call System.Void System.Object::.ctor()}	Mono.Cecil.Cil.Instruction
+		[13]	{IL_002d: nop}	Mono.Cecil.Cil.Instruction
+		[14]	{IL_002e: ldc.i4 1}	Mono.Cecil.Cil.Instruction
+		[15]	{IL_0033: ldc.i4 599}	Mono.Cecil.Cil.Instruction
+		[16]	{IL_0038: ldloca V_0}	Mono.Cecil.Cil.Instruction
+		[17]	{IL_003c: ldloca V_1}	Mono.Cecil.Cil.Instruction
+		[18]	{IL_0040: call System.Void nCrunch.TestRuntime.TestCoverageEventListener::NCrunchMarkTestLineStart(System.Int32,System.Int32,System.Int32&,System.UInt32&)}	Mono.Cecil.Cil.Instruction
+		[19]	{IL_0045: nop}	Mono.Cecil.Cil.Instruction
+		[20]	{IL_0046: ldc.i4 1}	Mono.Cecil.Cil.Instruction
+		[21]	{IL_004b: ldc.i4 600}	Mono.Cecil.Cil.Instruction
+		[22]	{IL_0050: ldloca V_0}	Mono.Cecil.Cil.Instruction
+		[23]	{IL_0054: ldloca V_1}	Mono.Cecil.Cil.Instruction
+		[24]	{IL_0058: call System.Void nCrunch.TestRuntime.TestCoverageEventListener::NCrunchMarkTestLineStart(System.Int32,System.Int32,System.Int32&,System.UInt32&)}	Mono.Cecil.Cil.Instruction
+		[25]	{IL_005d: nop}	Mono.Cecil.Cil.Instruction
+		[26]	{IL_005e: leave IL_0076}	Mono.Cecil.Cil.Instruction
+		[27]	{IL_0063: ldc.i4 1}	Mono.Cecil.Cil.Instruction
+		[28]	{IL_0068: ldloca V_0}	Mono.Cecil.Cil.Instruction
+		[29]	{IL_006c: ldloc V_1}	Mono.Cecil.Cil.Instruction
+		[30]	{IL_0070: call System.Void nCrunch.TestRuntime.TestCoverageEventListener::NCrunchExitMethod(System.Int32,System.Int32&,System.UInt32)}	Mono.Cecil.Cil.Instruction
+		[31]	{IL_0075: endfinally}	Mono.Cecil.Cil.Instruction
+		[32]	{IL_0076: ret}	Mono.Cecil.Cil.Instruction
```

Litee:
Maybe it would be easier just to always capture default public constructors (breaking change, of course)?

JakeGinnivan:
Yeah, lets just do that.

JakeGinnivan:
@citizenmatt thoughts? The breaking change seems to be the simplest way to fix this issue. Thought id reach out to you before merging

Litee:
Note that problem is not only about NCrunch - it may also happen with other tools - code coverage, AOP, etc

citizenmatt:
Yeah, it was fragile and rather nasty, but worked for my immediate need. Not upset to see it go.

However, I now can't think why I was removing an *empty* default constructor in the first place. Who cares if it's empty or not? I think the scenario I wanted to address was the noise when outputting a simple class - there's *always* a default constructor, it doesn't give me anything useful, so remove it.

Perhaps a better solution is to always hide a default constructor, unless there are other constructor overloads. A default constructor by itself is not useful, but knowing that there are three constructors, one of which has zero arguments is useful, especially if it's added/removed between versions. And this also hides the noise of auto-generated default constructors.

What do you think?

Litee:
If you hide default constructor then public -> private visibility change won't be noticed. In general, I believe that typically usage is to review changed lines via diff-based interface. Extra constructors would make little difference here.

JakeGinnivan:
Ok, we have 2 options:

1. Do not show default ctors, show private default ctor
2. Show all ctors including default

I am tempted by 1, just because it will be less of a breaking change for people.If people are using this for large code bases then it will cause a lot of tests to fail. Just seeing the number of change tests here.
But if we bump the major version then I am ok with that.

Litee:
Option 1 = "we show all constructors except those that are public && standalone && without arguments", correct? I find it contra-intuitive - unless you know about the problem it is hard to understand why private items are captured.

Options 2 - what I like about this option is that it is the most intuitive one. Also, if ApiApprover was updated against green tests then new snapshots can be approved straightaway. Agree with major version bump - it will send the right message to users.

Option 3 - introduce new optional "show/hide default constructor" flag with "hide" as a default.

citizenmatt:
I'm reasonably convinced to always show a public default constructor, so removing it will show as a breaking change.

JakeGinnivan:
Ditto. Will merge and prepare a 2.0 release with this change over the weekend

PR #9, opened 2015-02-15
Update README to reference a current example

MattDavies:


JakeGinnivan:
Maybe keep a simple example, then link to the more complex one?

MattDavies:
Thoughts on that?

JakeGinnivan:
Why `typeof(Application).Assembly.CodeBase.Replace("file:///", "")` rather than `typeof(Application).Assembly.Location` out of interest?

MattDavies:
None in particular, grabbed from an existing project. Swapped out.

JakeGinnivan:
Sweet, thanks!

PR #11, opened 2015-02-19
use App_Packages convention

### LINKED ISSUE #10: Move the approval code to the App_Packages directory
As per point 2 here http://nikcodes.com/2013/10/23/packaging-source-code-with-nuget/

should also adopt the `{package-id}.{package-version}` since this helps with updates

this also allows people to exclude your nuget source package from stylecop and resharper inspections

Want a PR?

SimonCropp:
fixes #10

i also removed ExampleApiApprovalTest from the package. i think doco is a better place for that and it will break for anyone not using xunit

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -7,6 +7,6 @@
     <ItemGroup>
       <Nuget Include="$(MSBuildProjectDirectory)\src\packages\NuGet.CommandLine.*\tools\NuGet.exe" />
     </ItemGroup>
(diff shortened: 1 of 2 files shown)


SimonCropp:
note the version is now hard coded in a diff single location.

SimonCropp:
also your website is case sensitive and doesnt do redirects

SimonCropp:
@JakeGinnivan bump :)

JakeGinnivan:
Been on a cruise ship so had no internet. Will probably sort when we get to Miami, otherwise early next week when back in London.

Sent from my Windows Phone
________________________________

SimonCropp:
LOL. no worries. have fun

Issue #17, opened 2016-02-12
Allow an attributes whitelist to be passed to the Generator on invocation


dee-see:
There are some attributes in my code that can change without inducing any breaking change to the API.

It would be nice to be able to give a list of attributes that would be combined with the `SkipAttributeNames` collection when the `ShouldIncludeAttribute` evaluates whether an attribute should be included or not.

If you agree with the feature I don't mind working on it and sending a pull request.

JakeGinnivan:
There are many changes which are non-breaking. Adding a method, or a new class for instance.

The idea of this is simply generate the public API, review and approve the changes with approval tests or shouldly's ShouldMatchApproved().

That said, if this is to reduce noise in general then sure I see no issues with this feature. I would probably use it for jetbrains annotations.

dee-see:
Yeah I misused the term breaking change. I meant that some attributes are not API changes at all, they are merely metadata.

I'm going to work on that and submit a pull request soon. Thanks!

danielmarbach:
Replaced by #45

PR #18, opened 2016-02-15
Allow an attributes whitelist to be passed to the Generator on invocation

dee-see:
I added the feature discussed in #17. I've introduced a design change to do it and I'd like to know your opinion about it.

Basically passing the excluded attributes list from `GetPublicApi` to `ShouldIncludeAttribute` was super messy and bloated the code for no good reason. I decided to change `PublicApiGenerator` to a concrete class (with a private constructor) and initialize the collection there.

It will make it easier for future similar new features to be added and it helps keeping the code readable, imo. The `GetPublicApi` static method is still the only public point of entry to the generator.

What are your thoughts on this?


danielmarbach:
Replaced by #44

danielmarbach:
@dee-see Thanks for all your effort. I reimplemented your proposal on top of current master. I couldn't find your email to make you a co-author. Sorry for that.

PR #20, opened 2016-03-16
Correction of a copy-paste error when processing Properties

dee-see:
GetMethod was passed to PopulateCustomAttributes instead of SetMethod


danielmarbach:
Thanks I'm pulling this in for v5

PR #23, opened 2016-11-04
PublicApiGenerator made source only package

danielmarbach:
- Bumped version to 5.0
- CecilEx.cs moved into ApiGenerator
- Renamed PublicApiGenerator to ApiGenerator to not clash with namespace

Upgraded NSB here https://github.com/Particular/NServiceBus/pull/4273

Thoughts @SimonCropp @andreasohlund @JakeGinnivan ?


thomaslevesque:
@JakeGinnivan @danielmarbach, what was the reason for this change?
I don't know what are the benefits of making it a source-only package, but I know the drawbacks... We use `PublicApiGenerator` as a binary package in FakeItEasy, and when I tried to upgrade to the latest version, I was greeted with dozens of StyleCop warnings, which I can't easily disable. Would it be possible to maintain a binary package as well as a source-only package?

danielmarbach:
The reason for this change is signing. With AppPackages the code is signed with the project settings of the target it is pulled in. Otherwise we'd need to sign the generator with a strong name for strong named test assemblies or provide both signed or unsigned versions. Resharper analysis allows to ignore all files in AppPackages folder. Sure stylecop can do this too?

thomaslevesque:
> Sure stylecop can do this too?

Probably, but I haven't figured out how to do it yet...

PR #25, opened 2016-11-15
Expose type filtering ability in PublicApiGenerator

### LINKED ISSUE #24: Expose type filtering in PublicApiGenerator.GetPublicApi
Currently it's possible to specify which types should be included, but there's no easy way to specify that specific types should be excluded. I see that internally, `GetPublicApi` calls `CreatePublicApiForAssembly`, which accepts a predicate on the types. Exposing this publicly would be very useful.

For completeness, here's my use case: in FakeItEasy, we embed Castle.Core using ILMerge; most of the types are internalized, but some need to be public. I'd like to exclude the Castle types from the approval tests, since they're exposed only "by accident" and are not really part of FakeItEasy's API.

thomaslevesque:
Fixes #24

### DIFF
--- a/src/PublicApiGenerator/PublicApiGenerator.cs
+++ b/src/PublicApiGenerator/PublicApiGenerator.cs
@@ -18,10 +18,10 @@ namespace PublicApiGenerator
 {
     public static class PublicApiGenerator
     {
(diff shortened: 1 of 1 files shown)


danielmarbach:
Thanks, not going to pull this in. Discussion and reasons are in the linked issue. Sorry that it took so long

PR #26, opened 2016-11-21
Async keyword is not part of the public API

### LINKED ISSUE #19: async keyword generates a different public API
If understood correctly, `async` keyword is just a syntax sugar and shouldn't change public API.
Right now it does. Here's a repro:

``` c#
public interface ITopology
{
    Task<bool> RunChecks();
}

public class A : ITopology 
{
    public Task<bool> RunChecks()
    {
        return Task.FromResult(true);
    }
}

public class B : ITopology
{
    public async Task<bool> RunChecks()
    {
        await Task.Delay(1); // mimicing some other call
        await Task.Delay(2); // mimicing some other call
        // logic to generate returned result based on invocations
        return true;
    }
}
```

Generate code is:

``` c#
public class A : ITopology
{
        public System.Threading.Tasks.Task<bool> RunChecks() { }
}   

public class B : ITopology
{
        public async System.Threading.Tasks.Task<bool> RunChecks() { }
}
```

danielmarbach:
Closes #19 

The async keyword should not be part of the public API at all.

The async keyword is an implementation detail since it is just syntactic sugar. From an API perspective the following APIs are equivalent:

- `async void` vs. `void`
- `async Task` vs `Task`
- `async Task<T>` vs `Task<T>`

From a behavior perspective you could argue that the runtime behavior from `void` to `async void` can change and therefore it could be a breaking change but from an API perspective both APIs are binary compatible. Since API approver's concern is only the public API the `async` keyword should be omitted.

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -5,7 +5,7 @@
     <Message Text=" ===========Packaging code===========" Importance="High" />
 
     <PropertyGroup>
-      <Version>5.0.0</Version>
+      <Version>6.0.0</Version>
     </PropertyGroup>
     <ItemGroup>
       <Nuget Include="$(MSBuildProjectDirectory)\src\packages\NuGet.CommandLine.*\tools\NuGet.exe" />
--- a/src/ApiApproverTests/Method_async.cs
+++ b/src/ApiApproverTests/Method_async.cs
@@ -21,6 +21,34 @@ namespace ApiApproverTests
         }
 
         [Fact]
+        public void Should_output_async_method()
+        {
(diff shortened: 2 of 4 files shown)


danielmarbach:
@seanfeldman feel free to review

danielmarbach:
Tested with NServiceBus https://github.com/Particular/NServiceBus/pull/4312/commits/154d2ee033936958d96721f2792e2d42a2da47e2#diff-1eb180a4f6d35cd0c53fde55c1a077b3L337

SeanFeldman:
👍

PR #30, opened 2017-05-11
Mono.Cecil.0.10.0-beta5

SimonCropp:
can we get a beta release with this change?


danielmarbach:
https://www.nuget.org/packages/ApiApprover/6.1.0-beta1

PR #31, opened 2017-05-18
Add support for contentFiles

bording:
@danielmarbach This updates the nuspec files to include the source files in the `contentFiles` folder as well as `content`. This let the packages work when installed in a project that uses NuGet's transitive restore mechanism, for example via PackageReference in the new SDK-style csproj format.

I've intentionally removed the version number from the `contentFiles` path because these will be refereenced from the NuGet package cache folder, which already includes the version number.


danielmarbach:
Merged and released as beta 2 Thx!

PR #35, opened 2017-09-19
Add AppDomain.CurrentDomain.BaseDirectory to assembly search path

GeertvanHorrik:
NUnit creates a lot of directories for the assemblies, meaning it's hard for the ApiGenerator to find dependencies. By adding *AppDomain.CurrentDomain.BaseDirectory* as search directory, it will have the *actual output directory* of the test project as search directory making sure that dependencies can be found when running using NUnit.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.CodeDom;
 using System.CodeDom.Compiler;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


danielmarbach:
@GeertvanHorrik thanks for bringing this up. Although I have commit rights here I'm still flying a bit blind when it comes to the build server. I'm trying to reach @JakeGinnivan for a few weeks now. As soon as I know more I'll see how we can get this in

GeertvanHorrik:
@danielmarbach is this something that is already done? Or should this still be added?

danielmarbach:
Superseded by #43

PR #41, opened 2017-11-17
Support .NET Standard 2.0

### LINKED ISSUE #28: Support for .NET Standard
.NET Standard projects are not currently supported

bording:
Now that Mono.Cecil 0.10.0-beta7 is out, there are no more blockers to being able to get this working on .NET Core and .NET Standard assemblies.

Because the APIApprover project has a dependency on ApprovalTests, I have not updated it to also have a netstandard2.0 target. We'd need to chose another approval tests project to use, for example [Assent](https://github.com/droyad/Assent) which works with .NET Standard if we want to do that as well.

In addition, I've updated a number of things here to fix up some packaging issues, and have generally cleaned things up a bit.

I've also fixed the AppVeyor script so that it builds, tests pass, and the NuGet packages are available from the artifacts tab.

Fixes #28


bording:
This is currently resulting in a 6.4.1 version. I'm assuming we'd probably want to bump to 6.5 for this change. Should I do that in this PR?

danielmarbach:
Yes please. Will review later today

PR #42, opened 2018-03-07
Fix potential NullReferenceException for unresolved attribute types

GeertvanHorrik:


danielmarbach:
Thanks @GeertvanHorrik 

Have you observed the NullRef in practice in your code bases? If yes could you add here the code that triggered it so that we can reproduce it with a unit test? Happy to write that test for you.

GeertvanHorrik:
@danielmarbach yep, it was when using attributes from a (temporary) Fody weaver attribute (in this case Catel.Fody, I won't bother you with the details).

See https://github.com/WildGums/Orc.SupportPackage/commit/a5c8f62e33206f896793c4d085c66defb6ddfa98

danielmarbach:
Merged and release 6.5.1 (will take a while to show up on NuGate ;) )

GeertvanHorrik:
Thanks! I had to wait 2 hours this morning for a nuget package to be available in the clients. Thanks anyway for helping out!

PR #48, opened 2018-05-03
Fix whitelisting System namespaces

ryanwersal:
I noticed that while whitelisting works for "Microsoft" prefixes it wasn't allowing whitelisting to override the filtering of "System" prefixes.


danielmarbach:
Thanks for the contribution. I'll try to release it azät latest on Friday (currently I'm on vacation)

ryanwersal:
No problem, thanks for reviewing and have a great vacation!

danielmarbach:
Will give it another closer look later. Cell phone reviews are not always the best ones 😂 if it requires cleanup I will do it while merging

PR #49, opened 2018-06-27
Include operators in api generation

ryanwersal:
I noticed that operators are not currently included in the generated public api surface whereas I would expect operators to be included. This was the result of `IsSpecialName` returning `true` for operators.

Unfortunately there doesn't appear to be a good way to tell whether this is a "special name" that should be included or not (such as the case for operators). I confess I'm not sure what all falls under the "special name" category. The best differentiator I could find [was sources claiming that all operator names are formatted as "op_*"](https://stackoverflow.com/a/3016653/558700) so I leveraged that for determining if a given method is an operator. I'm very open to alternatives that can be more precise without needing to rely on string comparisons.

I'm willing to write more tests but I wasn't sure what would be overkill. I tried to mirror the `Method_*.cs` test files to a degree.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -510,17 +510,50 @@ namespace PublicApiGenerator
             typeDeclaration.Members.Add(method);
         }
 
+        static readonly IDictionary<string, string> OperatorNameMap = new Dictionary<string, string>
+        {
+            { "op_Addition", "+" },
+            { "op_UnaryPlus", "+" },
+            { "op_Subtraction", "-" },
+            { "op_UnaryNegation", "-" },
+            { "op_Multiply", "*" },
+            { "op_Division", "/" },
+            { "op_Modulus", "%" },
+            { "op_Increment", "++" },
+            { "op_Decrement", "--" },
+            { "op_OnesComplement", "~" },
+            { "op_LogicalNot", "!" },
+            { "op_BitwiseAnd", "&" },
+            { "op_BitwiseOr", "|" },
+            { "op_ExclusiveOr", "^" },
+            { "op_LeftShift", "<<" },
+            { "op_RightShift", ">>" },
+            { "op_Equality", "==" },
+            { "op_Inequality", "!=" },
+            { "op_GreaterThan", ">" },
+            { "op_GreaterThanOrEqual", ">=" },
+            { "op_LessThan", "<" },
+            { "op_LessThanOrEqual", "<=" }
+        };
+
         static void AddMethodToTypeDeclaration(CodeTypeDeclaration typeDeclaration, MethodDefinition member, HashSet<string> excludeAttributes)
         {
+            if (member.IsAssembly || member.IsPrivate) return;
+
             var isOperator = member.Name.StartsWith("op_");
-            if (member.IsAssembly || member.IsPrivate || (member.IsSpecialName && !isOperator))
-                return;
+            if (member.IsSpecialName && !isOperator) return;
+
+            var memberName = member.Name;
+            if (isOperator && OperatorNameMap.ContainsKey(memberName))
+            {
+                memberName = OperatorNameMap[memberName];
+            }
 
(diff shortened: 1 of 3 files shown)


src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -512,7 +512,8 @@ static void AddCtorToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Method
 
         static void AddMethodToTypeDeclaration(CodeTypeDeclaration typeDeclaration, MethodDefinition member, HashSet<string> excludeAttributes)
         {
-            if (member.IsAssembly || member.IsPrivate || member.IsSpecialName)
+            var isOperator = member.Name.StartsWith("op_");
```
danielmarbach: Could we inline this so that it is only evaluated with IsSpecialName?

src/PublicApiGeneratorTests/Operator_order.cs
```diff
@@ -0,0 +1,48 @@
...
+        public void Should_show_custom_operators()
+        {
+            AssertPublicApi<ClassWithUnaryOperators>(
+@"namespace PublicApiGeneratorTests.Examples
+{
+    public class ClassWithUnaryOperators
+    {
+        public ClassWithUnaryOperators() { }
+        public static PublicApiGeneratorTests.Examples.ClassWithUnaryOperators op_Addition(PublicApiGeneratorTests.Examples.ClassWithUnaryOperators first, PublicApiGeneratorTests.Examples.ClassWithUnaryOperators second) { }
+        public static PublicApiGeneratorTests.Examples.ClassWithUnaryOperators op_Decrement(PublicApiGeneratorTests.Examples.ClassWithUnaryOperators first) { }
```
danielmarbach: Would it make sense to have a translation table that translates comnon well known ops into the corresponding operator code? Would be a bit more readable in the output I think

danielmarbach:
Awesome stuff. Made a few comments

src/PublicApiGeneratorTests/Operator_order.cs
ryanwersal: I agree. I looked into why the full type name was being generated but, likely due to operators being static, it appears they are seen as "unnested" types which causes the `FullName` to be used instead of just the specific type's name.

Handling the operators in a more special manner would definitely allow me to more cleanly format them (since the problematic area only knows the `Type` without any context of its use). Do you have a preference of such output? To riff on something:

```
public static ClassWithUnaryOperator +(ClassWithUnaryOperator first, ClassWithUnaryOperator second) { }
```

Or are you looking to further condense the signature?

src/PublicApiGenerator/ApiGenerator.cs
ryanwersal: Definitely.

danielmarbach:
I think it would be good enough to handle known operators except implicit casts as their code represention like your sample shows. Doesnt have to be fancy. We could always improve in a next major after releasing the first initial version (will also be a major)

danielmarbach:
@ryanwersal do you need help to get this going?

ryanwersal:
@danielmarbach Sorry, I admit I somewhat forgot about this... I'll look to getting it updated over the weekend.

danielmarbach:
No worries. Good stuff takes time 💃

ryanwersal:
Apologies for the near month (!!!) delay but I have the PR updated with a mapping as requested. 

I did the operators I could find on the [Overloadable Operators](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/overloadable-operators) documentation except for the true/false operators as I found it a bit ambiguous in the final output (since it looks so similar to a boolean value/output).

Let me know if you'd like further modifications and/or add more tests. Thanks!

danielmarbach:
@ryanwersal scandalous :) :trollface: 

Don't be so hard to yourself. You did awesome work here and every contribution is very much appreciated!

The packages are being indexed on nuget. In a few minutes your work is on the interwebs

Thanks again!

PR #52, opened 2018-09-15
fix typos and use pattern matching

SimonCropp:


danielmarbach:
Thanks Simon. Given that we should as sources I'm not sure we can already assume a language version that supports pattern matching. What are your thoughts?

SimonCropp:
Yeah i didnt think of that. So those features have been out for over 12 months, imo i would document it as a min requirement and move forward. But ultimately you have to handle the people complaining, so up to you :)

Have u considered shipping as a dll. I know i would prefer that method of consumption

SimonCropp:
let me know if u want a new PR with just the spelling

danielmarbach:
@SimonCropp the reason why we made this source only was this

https://github.com/JakeGinnivan/ApiApprover/pull/23#issuecomment-269939420

Any thoughts on that? I mean we could switch back to binary but then we probably need to sign the package.

danielmarbach:
@SimonCropp thank you for the contributions. I split the pattern matching out for now and committed

https://github.com/JakeGinnivan/ApiApprover/commit/703b483c35cd34c88e86db3aea796c4a6c94fae2
https://github.com/JakeGinnivan/ApiApprover/commit/b994b222a3377b9de6561db634dc18483453cd48

PR #53, opened 2019-01-06
Update Mono.Cecil to 0.10.1 to allow Embedded debug type

glennawatson:
At the moment 0.10 of Cecil only supports portable PDB and full PDBs

They added support for the embedded files in v0.10.1 version of Mono.Cecil

Updating the reference to allow for this.

The current workaround is easy in terms of I can just update my project but thought others might benefit from this.

Thanks :)


danielmarbach:
Thanks for the contribution. Released as 8.1.0!

PR #58, opened 2019-02-12
 Use StringComparer.Ordinal instead of linguistic culture aware ordering which can lead to different results depending on the environment

### LINKED ISSUE #56: Ordinal ordering of classes, interfaces, members and fields
Currently, default string ordering is used which can lead to inconsistent ordering depending on the environment in which the API diff is generated. Ideally ordering of classes, interfaces, members and fields should be done using the StringComparison.Ordinal ordering rules.

danielmarbach:
Closes #56


bording:
LGTM. This should definitely solve the problem I ran into!

PR #60, opened 2019-02-12
Switch to binaries, SignAssemblies, ApiApprover netstandard2.0

danielmarbach:
This switches to binaries and moves away from source shipping
ApiApprover will become netstandard2.0 (only due to ApprovalTests not supporting multiple target frameworks)
Assemblies strong named


danielmarbach:
@SimonCropp @bording @thomaslevesque @andreasohlund I'm switching to binaries only. I see no reason to keep PublicApiApprover source package. Any objections or reasons why this might not be a good idea?

SimonCropp:
Go for it

src/ApiApprover.sln
```diff
@@ -15,6 +15,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		..\build.ps1 = ..\build.ps1
 		Directory.Build.props = Directory.Build.props
 		..\GitVersion.yml = ..\GitVersion.yml
+		GitVersionWorkaround.targets = GitVersionWorkaround.targets
```
bording: This should be removed.

src/ApiApprover.sln
```diff
@@ -15,6 +15,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
...
+		..\PublicApiGenerator.pfx = ..\PublicApiGenerator.pfx
```
bording: This should be changed to match the new file name, or maybe removed.

src/ApiApprover/ApiApprover.csproj
```diff
@@ -1,13 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net40</TargetFramework>
```
bording: As long as you're keeping this package around, I'd suggest multi-targeting it as `<TargetFrameworks>netstandard2.0;net40</TargetFrameworks>` as well.

bording:
I also suggest two other changes;
- Update Mono.Cecil to latest since it fixes an attribute ordering problem
- Update PublicApiGeneratorTests to be `netcoreapp2.1` instead of `netcoreapp2.0`

bording:
> I'm switching to binaries only. I see no reason to keep PublicApiApprover source package. Any objections or reasons why this might not be a good idea?

As long as you're going to be putting this out a new major, it seems fine to me. I've never found a reason to want this as a source package since I only ever reference it from test projects anyway.

src/ApiApprover.sln
danielmarbach: I did remove this on master. Let me check

src/ApiApprover/ApiApprover.csproj
danielmarbach: Like mentioned in the description I can't multi-target because ApprovalTests only supports netstandard2.0

bording: Ah, I didn't notice that. Then ApprovalTests is doing the wrong thing.

At this point, you should always be multi-targeting to at least `net461;netstandard2.0`. Just targeting `netstandard2.0` is basically the wrong thing to do because support for .NET Standard 2.0 is broken in various ways in `net461` - `net471`.

danielmarbach:
I think I'll add an obsolete with warning to the ApiApprover and remove it in the next major version. It is time to let it die

src/ApiApprover/PublicApiApprover.cs
```diff
@@ -1,12 +1,15 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using System.Reflection;
 using ApprovalTests;
 using ApprovalTests.Namers;
 
 namespace ApiApprover
 {
+    [Obsolete("PublicApiApprover will be removed in the next major version. Either use the `ApiGenerator` directly or copy https://github.com/JakeGinnivan/ApiApprover/blob/master/src/ApiApprover/PublicApiApprover.cs into your repository if you plan to continue to use ApprovalTests.", false)]
```
danielmarbach: @bording spell checking would be highly appreciated ;)

bording: This seems a bit strange to me, because you'll actually just stop publishing the ApiApprover package, right? The PublicApiGenerator package would be the only one still being published?

If I'm understanding the plan correctly, then these feel misplaced since you'd actually want to tell people to migrate to a new package instead.

bording:
Based on https://github.com/JakeGinnivan/ApiApprover/pull/60#discussion_r256162089, I'd still argue that ApiApprover should be `net461;netstandard2.0` even if ApprovalTests isn't.

danielmarbach:
@bording https://github.com/approvals/ApprovalTests.Net/pull/260 ;)

src/ApiApprover/PublicApiApprover.cs
danielmarbach: Changed the message

SimonCropp:
Do i have a use case for net461?

danielmarbach:
@SimonCropp based on https://twitter.com/terrajobst/status/1031999730320986112?lang=de I read the conclusion is multi-targeting is pretty much the way to go

danielmarbach:
See also https://twitter.com/davidfowl/status/1034105392475791361

src/ApiApprover/PublicApiApprover.cs
```diff
@@ -1,12 +1,15 @@
-﻿using System.IO;
+﻿using System;
+using System.IO;
 using System.Reflection;
 using ApprovalTests;
 using ApprovalTests.Namers;
 
 namespace ApiApprover
 {
+    [Obsolete("The package `ApiApprover` will be removed in the next major version. Install the package `PublicApiGenerator` and use the `ApiGenerator` directly or copy https://github.com/JakeGinnivan/ApiApprover/blob/master/src/ApiApprover/PublicApiApprover.cs into your repository if you plan to continue to use ApprovalTests in combination with the `ApiGenerator`.", false)]
```
bording:
```suggestion
    [Obsolete("The `ApiApprover` package will be removed in the next major version. Install the `PublicApiGenerator` package and use the `ApiGenerator` class directly instead.", false)]
```

I removed the link to the code because it linked to the master branch, but that code will eventually be removed when ApiApprover is deleted.

If you think that's important to keep, perhaps the example could be added to the README instead, and some sort of sentence saying "To see an example of how to use PublicApiGenerator with ApprovalTests, see the README."

thomaslevesque:
> Any objections or reasons why this might not be a good idea?

No objection from me!

PR #66, opened 2019-07-15
Output protected internal access modifier as protected

jonorossi:
At the moment members marked with the `protected internal` access modifier are output inconsistently:
- Fields get no modifier
- Methods and properties get `protected internal`

Since the `internal` part of `protected internal` doesn't apply to the public API, I think all `protected internal` members should be output like they are just marked `protected`.

I've provided a fix and added/updated unit tests. Let me know if there is anything else you'd like me to do.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -620,16 +620,12 @@ namespace PublicApiGenerator
         static MemberAttributes GetMethodAttributes(MethodDefinition method)
         {
             MemberAttributes access = 0;
-            if (method.IsFamily)
+            if (method.IsFamily || method.IsFamilyOrAssembly)
                 access = MemberAttributes.Family;
             if (method.IsPublic)
                 access = MemberAttributes.Public;
             if (method.IsAssembly)
                 access = MemberAttributes.Assembly;
-            if (method.IsFamilyAndAssembly)
-                access = MemberAttributes.FamilyAndAssembly;
-            if (method.IsFamilyOrAssembly)
-                access = MemberAttributes.FamilyOrAssembly;
 
             MemberAttributes scope = 0;
             if (method.IsStatic)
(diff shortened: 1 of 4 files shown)


danielmarbach:
Currently on vacation. I'll look once I have more time. Might take a while sorry

jonorossi:
Thanks, have a great holiday.

bording:
Given that `protected internal` is a weird combo access modifier, I think these changes make sense. The fact that the member is `internal` as well isn't really relevant for the API surface.

If I had something declared `protected internal` and then changed it to `protected`, would that be a breaking change in any way? I don't think it would be.

Assuming that is correct, then I'm 👍 for the change. 

If does turn out to be a breaking change, then `protected internal` should be displayed in the output.

danielmarbach:
Could anyone be relying on the fact that we dump InternalsVisibleToAttributed and the internal modifier for protected and then consider this the public API surface within those boundaries?

bording:
> Could anyone be relying on the fact that we dump InternalsVisibleToAttributed and the internal modifier for protected and then consider this the public API surface within those boundaries?

In order for that to really work, PublicApiGenerator would need to be showing all of the `internal` members, not just the `protected internal` ones, so I find it difficult to believe this is a valid use case currently.

The `InternalsVisibleToAttribute` already implies some very tight coupling, so I'd hate to think someone is relying on an API approval test to not break the assembly listed in the attribute. If the assemblies aren't in the same project with a direct project reference, you're likely doing something wrong/weird in the first place.

However, If there was some way to optionally display `internal` members as well, then I would be more inclined to include the `internal` portion of `protected internal` when in that mode. Then perhaps the idea of "let's document the internal stuff as well because there is an `InternalsVisibleToAttribute` would be more feasible."

jonorossi:
> Could anyone be relying on the fact that we dump InternalsVisibleToAttributed and the internal modifier for protected and then consider this the public API surface within those boundaries?

As @bording already said, `internal` members currently aren't output.

In C# `protected internal` is `FamilyOrAssembly`:
> The type or member can be accessed by any code in the assembly in which it is declared, or from within a derived class in another assembly.
> (https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers)

`FamilyAndAssembly` isn't possible in C#:
> The Visual Basic and C# languages cannot define methods with MethodAttributes.FamANDAssem visibility; that access level appears only in the C++ example.
> (https://docs.microsoft.com/en-us/dotnet/api/system.reflection.methodbase.isfamilyorassembly)

danielmarbach:
Thanks, @jonorossi for this contribution! Will release this after my vacation in roughly two weeks

PR #72, opened 2019-08-02
WIP: Fix out parameter generic types

jonorossi:
I just upgraded to 9.2.0 in Castle Core, thanks for the release.

After the changes in #69 for generic constraints, I have now noticed that our `out` parameters didn't have their generic parameters specified before (e.g. `IEnumerable<>` was the parameter type), but are now outputting the generic constraint name (or `!n`) rather than being populated with the type from the parameter.

```csharp
// Input:
public void Method(out GenericType<int> value) { }
public void Method(out IEnumerable<int> value) { }
public void Method(out GenericType<ComplexType> value) { }

// Output:
public void Method(out GenericType<T> value) { }
public void Method(out IEnumerable<!0> value) { }
public void Method(out GenericType<T> value) { }
```

I've included the unit tests I wrote in this draft pull request so people are aware of the defect. I tried quickly to fix the problem but couldn't see how to fix it. I'll return to it again next week, if you have any tips on what you think might be the cause happy to listen.

### DIFF
--- a/src/PublicApiGeneratorTests/Method_parameters.cs
+++ b/src/PublicApiGeneratorTests/Method_parameters.cs
@@ -174,6 +174,36 @@ namespace PublicApiGeneratorTests
         }
 
         [Fact]
+        public void Should_output_out_parameter_with_generic_type()
+        {
+            AssertPublicApi<MethodWithOutGenericTypeParameter>(
+@"namespace PublicApiGeneratorTests.Examples
+{
+    public class MethodWithOutGenericTypeParameter
+    {
+        public MethodWithOutGenericTypeParameter() { }
+        public void Method(out PublicApiGeneratorTests.Examples.GenericType<int> value) { }
+    }
+}");
+        }
+
+        [Fact]
+        public void Should_output_out_parameter_fully_qualified_type_name_for_generic_parameter()
+        {
+            AssertPublicApi<MethodWithOutGenericTypeOfComplexTypeParameter>(
+@"namespace PublicApiGeneratorTests.Examples
+{
+    public class MethodWithOutGenericTypeOfComplexTypeParameter
+    {
+        public MethodWithOutGenericTypeOfComplexTypeParameter() { }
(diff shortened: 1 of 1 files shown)


danielmarbach:
Interestingly when I undo the changes done in #69 those tests don't pass either. Investigating. So that was actually never really supported it seems. The actual output before the changes was

```
namespace PublicApiGeneratorTests.Examples
{
    public class MethodWithOutGenericTypeParameterEnumerable
    {
        public MethodWithOutGenericTypeParameterEnumerable() { }
        public void Method(out System.Collections.Generic.IEnumerable<> value) { }
    }
}
```

after the changes

```
namespace PublicApiGeneratorTests.Examples
{
    public class MethodWithOutGenericTypeParameterEnumerable
    {
        public MethodWithOutGenericTypeParameterEnumerable() { }
        public void Method(out System.Collections.Generic.IEnumerable<!0> value) { }
    }
}
```

jonorossi:
> So that was actually never really supported it seems. The actual output before the changes was

Yer, I mentioned the previous behaviour in the pull request description. We had `out IEnumerable<>` before, we've gone from unnoticed broken to noticed broken 😉, so time to fix it.

danielmarbach:
Yeah sorry I was just clarifying what I understood to be sure I understood correctly

danielmarbach:
Would you mind testing out https://github.com/JakeGinnivan/ApiApprover/pull/74 with Castle?

jonorossi:
> Would you mind testing out #74 with Castle?

That fixed it! Works for both ref and out closed and open generic parameters.

danielmarbach:
Merged https://github.com/JakeGinnivan/ApiApprover/pull/74

danielmarbach:
9.3.0 is up on nuget. While take a while until it is indexed. Thanks man for providing the tests. You made my life so much easier!

jonorossi:
@danielmarbach cheers. Just finished upgrading Castle Core, 9.3.0 is now in its master.

No problem, I know personally how much a set of unit tests help. Thanks for the quick turnaround, was expecting to have to pick this up next week.

Issue #75, opened 2019-08-09
Multiline text in attributes differs on Windows and Linux


sungam3r:
I found that when using multi-line text in attributes, our tests fail with on CI, although locally everything is fine. The reason is line endings although it seems that ApiApprover already has built-in tools to bypass it. I use `ObsoleteAttribute` with detailed description which takes several lines. So I have `\r\n` literal in the middle of the line in approved file on Windows machine but on CI build server I get an error with `\n` literal. I use Shouldly - `ShouldMatchApproved` method.

I went through the ApiApprover code as a debugger (normalizing strings, etc.), but I didn’t figure out what should be done in this case.

sungam3r:
![image](https://user-images.githubusercontent.com/21261007/62862524-bb3c6500-bd0e-11e9-825c-bf3330ac7a4a.png)
@danielmarbach What does it mean?

Because of that normalization [this replace](https://github.com/shouldly/shouldly/blob/master/src/Shouldly/Internals/StringHelpers.cs#L178) in Shouldly does not work:
![image](https://user-images.githubusercontent.com/21261007/62862583-e8891300-bd0e-11e9-81d3-480096ad7291.png)

Before normalization:
![image](https://user-images.githubusercontent.com/21261007/62862690-46b5f600-bd0f-11e9-9ed0-c746dd4b6487.png)

After normalization:
![image](https://user-images.githubusercontent.com/21261007/62862720-5e8d7a00-bd0f-11e9-8938-e49bfdb1bb5e.png)

danielmarbach:
I have no idea honestly why this code is there. Paging @JakeGinnivan maybe he recalls. 

That being said can't shouldly ignore those cases?

sungam3r:
Shouldly **can** ignore line endings, but the problem is that `\\r\\n` **is no longer** a line ending.

danielmarbach:
So the code does this

```
        static string NormaliseLineEndings(string value)
        {
            return Regex.Replace(value, @"\r\n|\n\r|\r|\n", Environment.NewLine);
        }
```

which replaces it with the platform-specific variant NewLine. So that means you'd have to Replace Newline with `\r\n` before you pass it to shoudly and then you'd be safe correct?

sungam3r:
[...]

PR #76, opened 2019-08-09
test for #75

sungam3r:
@danielmarbach This is for a demonstration #75 . This test passes on my Windows machine, but I think that there should not be any carriage return characters at all.


sungam3r:
As I expected, the test failed.

danielmarbach:
Thanks for raising the PR. not going to pursue with this one as outlined in #75

PR #88, opened 2019-09-17
Add support for default keyword for non primitive structs

sungam3r:
I noticed that ApiApprover generates `CancellationToken token = null` for such common API like `public static Task DoSomething(CancellationToken token = default)`. It's obvious that `null` is invalid default value for struct.


sungam3r:
Is anything else required for merging?

sungam3r:
@danielmarbach Could you please push new package to nuget.org?

danielmarbach:
10.0.0-beta1 is on nuget

PR #89, opened 2019-09-18
Remove net40

danielmarbach:
Removes support for NET4.0 and switches to .NET Standard only


src/PublicApiGenerator/PublicApiGenerator.csproj
```diff
@@ -16,5 +16,5 @@
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
sungam3r: Now you can remove this condition too.

danielmarbach: Good catch. Forgot to save locally

PR #94, opened 2019-10-14
Weird readonly Func

sungam3r:
Fixes issue when 
`public readonly System.Func<a,b,c,d> Some;`
 become
 `public readonly System.Func<a Some;`

@danielmarbach I stumbled upon this by accident. I localized the problem - it happens if the control falls into the `ModifyCodeTypeReference` method. If `typeReference` has any `TypeArguments` then after call to `new CodeTypeReference(modifier + " " + provider.GetTypeOutput(typeReference))` there are 0 `TypeArguments` in newly created `CodeTypeReference`. Well, then comes complete nonsense.
If you call a constructor that accepts second argument - `params CodeTypeReference[] typeArguments`, then the output changes (`int` -> `System.Int32`, `string` -> `System.String` and so on) and some tests fail.

I used a simple if condition and made all the tests work. But I **do not like it**. This is a temporary solution just to PR. I have no experience with CodeDom. I think you better know what to do in this case. Perhaps everything is simple and we should switch to a new constructor without a condition. Please see how to fix it.

### DIFF
--- a/src/PublicApiGeneratorTests/Field_types.cs
+++ b/src/PublicApiGeneratorTests/Field_types.cs
@@ -86,6 +86,7 @@ namespace PublicApiGeneratorTests
     public class FieldWithFunc
     {
         public readonly System.Func<string, string, string, System.Collections.Generic.IEnumerable<string>, string> FuncField;
+        public readonly System.Func<string, string, string, string> FuncField2;
         public FieldWithFunc() { }
     }
 }");
@@ -124,6 +125,7 @@ namespace PublicApiGeneratorTests
         public class FieldWithFunc
         {
             public readonly Func<string, string, string, IEnumerable<string>, string> FuncField = (a, b, c, d) => null;
+            public readonly Func<string, string, string, string> FuncField2;
         }
     }
     // ReSharper restore UnusedMember.Global
--- a/src/PublicApiGeneratorTests/Field_types.cs
+++ b/src/PublicApiGeneratorTests/Field_types.cs
@@ -1,4 +1,6 @@
 ﻿using PublicApiGeneratorTests.Examples;
+using System;
+using System.Collections.Generic;
 using Xunit;
 
 namespace PublicApiGeneratorTests
@@ -74,6 +76,20 @@ namespace PublicApiGeneratorTests
     }
 }");
         }
+
+        [Fact]
+        public void Should_output_readonly_func_with_type_arguments()
+        {
+            AssertPublicApi<FieldWithFunc>(
+@"namespace PublicApiGeneratorTests.Examples
+{
+    public class FieldWithFunc
+    {
+        public readonly System.Func<string, string, string, System.Collections.Generic.IEnumerable<string>, string> FuncField;
+        public FieldWithFunc() { }
+    }
+}");
+        }
     }
 
     // ReSharper disable ClassNeverInstantiated.Global
@@ -103,6 +119,11 @@ namespace PublicApiGeneratorTests
         public class FieldWithMultipleGenericTypeParameters
         {
             public GenericTypeExtra<int, string, ComplexType> Field;
+        }
+
+        public class FieldWithFunc
+        {
+            public readonly Func<string, string, string, IEnumerable<string>, string> FuncField = (a, b, c, d) => null;
         }
     }
     // ReSharper restore UnusedMember.Global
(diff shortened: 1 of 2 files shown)


src/PublicApiGeneratorTests/Field_types.cs
```diff
@@ -72,6 +74,20 @@ public class FieldWithMultipleGenericTypeParameters
...
+
+        [Fact]
+        public void Should_output_readonly_func_with_type_arguments()
+        {
+            AssertPublicApi<FieldWithFunc>(
+@"namespace PublicApiGeneratorTests.Examples
+{
+    public class FieldWithFunc
+    {
+        public readonly System.Func<string, string, string, System.Collections.Generic.IEnumerable<string>, string> FuncField;
```
sungam3r: If remove `readonly` then everything work fine without changes in this PR.

src/PublicApiGeneratorTests/Field_types.cs
```diff
@@ -104,6 +120,11 @@ public class FieldWithMultipleGenericTypeParameters
         {
             public GenericTypeExtra<int, string, ComplexType> Field;
         }
+
+        public class FieldWithFunc
+        {
+            public readonly Func<string, string, string, IEnumerable<string>, string> FuncField = (a, b, c, d) => null;
```
sungam3r: If remove readonly then everything work fine without changes in this PR.

src/PublicApiGeneratorTests/Field_types.cs
```diff
@@ -104,6 +121,12 @@ public class FieldWithMultipleGenericTypeParameters
...
+            public readonly Func<string, string, string, string> FuncField2;
```
sungam3r: Note - this form of Func (without IEnumerable) works fine without changes in this PR **even with readonly**!

danielmarbach:
The problem is that CodeTypeReference does the following

 ```
private string RipOffAssemblyInformationFromTypeName(string typeName)
    {
      int startIndex = 0;
      int index1 = typeName.Length - 1;
      string str = typeName;
      while (startIndex < typeName.Length && char.IsWhiteSpace(typeName[startIndex]))
        ++startIndex;
      while (index1 >= 0 && char.IsWhiteSpace(typeName[index1]))
        --index1;
      if (startIndex < index1)
      {
        if (typeName[startIndex] == '[' && typeName[index1] == ']')
        {
          ++startIndex;
          --index1;
        }
        if (typeName[index1] != ']')
        {
          int num = 0;
          for (int index2 = index1; index2 >= startIndex; --index2)
          {
            if (typeName[index2] == ',')
            {
              ++num;
              if (num == 4)
              {
                str = typeName.Substring(startIndex, index2 - startIndex);
                break;
              }
            }
          }
        }
      }
      return str;
    }
```

which assumes that after four `,` are found from the end it can cut off. 💥 

That's why 

```
public readonly Func<string, string, string, string> FuncField2
```

works but 

```
Func<string, string, string, IEnumerable<string>, string>
```

does not

**Comment by danielmarbach** [comment]
This would fix it as well because it would bypass the cut off logic

```
        static CodeTypeReference ModifyCodeTypeReference(CodeTypeReference typeReference, string modifier)
        {
            using (var provider = new CSharpCodeProvider())
            {
                var typeOutput = modifier + " " + provider.GetTypeOutput(typeReference);
                var codeTypeReference = new CodeTypeReference(typeOutput);

                // for more than 4 generic params the RipOffAssemblyInformationFromTypeName of CodeTypeReferences starts breaking
                if (typeReference.TypeArguments.Count > 4)
                {
                    typeof(CodeTypeReference).GetField("_baseType", BindingFlags.NonPublic |
                                                                    BindingFlags.Instance)
                        ?.SetValue(codeTypeReference, typeOutput);
                }
                return codeTypeReference;
            }
        }
```

but honestly this is even nastier than your solution your solution works even with more parameter (although mine as well). 

The only benefit I see from the reflection hack would be that it is clearly visible why such weirdness is done

sungam3r:
> The problem is that CodeTypeReference does the following

Triple facespalm... So what do we do in this situation? Which option do we choose?

danielmarbach:
I pulled it in with a slight tweak on the doco. Thanks mate!

sungam3r:
Will you pubilsh a new nuget?

danielmarbach:
It is currently slow cooking in sous-vide by 56 degrees celcius to create the right medium rare inner body, when it is done I will use the flame thrower to give it a crisp outer texture to caramelize the code ;)

Will be out shortly

danielmarbach:
It's cooked and ready

sungam3r:
Don't see yet
https://www.nuget.org/packages/ApiApprover/

danielmarbach:
https://www.nuget.org/packages/PublicApiGenerator/10.0.0-beta2 😁

sungam3r:
completely forgot that the name is different

PR #98, opened 2019-10-20
Projects with /nullable:enable can't silence these warnings

### LINKED ISSUE #97: Including ApiGenerator.cs in projects with <Nullable>enable</Nullable> causes unsuppressible warnings
The PublicApiGenerator package includes a file as source which results in warnings when included in any project with `<Nullable>enable</Nullable>`:

image

GlobalSuppressions.cs is not an option because the warnings are coming from the C# compiler itself. The only recourse is to add pragmas to the source file itself to silence a CS8600, CS8603, and CS8625 (so far). Changes to the source file itself can only be done by the package publisher.

jnm2:
Fixes #97.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -1,3 +1,10 @@
+// In case this file is included in a project with the <Nullable>enable</Nullable> csproj property (which sets the
+// /nullable:enable compiler option):
+
+#pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
+#pragma warning disable CS8603 // Possible null reference return.
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+
 using System;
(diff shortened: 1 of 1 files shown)


jnm2:
This was already addressed in 9.x by not adding source files to the project.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -1,3 +1,10 @@
+// In case this file is included in a project with the <Nullable>enable</Nullable> csproj property (which sets the
+// /nullable:enable compiler option):
```
sharwell: This comment should have included a `<auto-generated/>` line if the file is meant to be imported effectively as generated code (outside the normal developer workflow). Generated files _default_ to `#nullable disable` even in projects with nullable reference types enabled, so this header would have corrected the warnings you saw.

jnm2: Excellent, thank you!

PR #110, opened 2019-11-02
Fix visibility issues with events, nested types, and 'private protected'

### LINKED ISSUE #109: Private protected methods are not API
C# 7.2 exposed the `famandassem` visibility that .NET binaries have always supported: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/private-protected

(`protected internal` compiles to `famorassem`.)

For:

```cs
public class ClassWithPrivateProtectedMethod
{
    private protected void DoSomething()
    {
    }
}
```

ApiApprover generates:

```cs
namespace PublicApiGeneratorTests.Examples
{
    public class ClassWithPrivateProtectedMethod
    {
        public ClassWithPrivateProtectedMethod() { }
        void DoSomething() { }
    }
}
```

But `DoSomething` should not be visible.

PR incoming.

jnm2:
Fixes #109. 

I set out to fix `famandassem` members which should not be shown. They have been able to be created in .NET assemblies since the beginning.

While writing tests, I discovered that tests were missing for the visibility of events and nested `protected internal` types. I added them and they failed, so this PR fixes two additional issues:

- Events were being included and shown as `public` regardless of visibility, including private ones.
- Nested `protected internal` types were excluded even though they appear as `protected` types from outside the assembly.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -107,13 +107,14 @@ namespace PublicApiGenerator
             return !m.IsCompilerGenerated() && !IsDotNetTypeMember(m, whitelistedNamespacePrefixes) && !(m is FieldDefinition);
         }
 
-        static bool ShouldIncludeMethod(Mono.Cecil.MethodAttributes methodAttributes)
+        static bool ShouldIncludeMember(MemberAttributes memberAttributes)
         {
-            switch (methodAttributes & Mono.Cecil.MethodAttributes.MemberAccessMask)
+            switch (memberAttributes & MemberAttributes.AccessMask)
             {
-                case Mono.Cecil.MethodAttributes.Private:
-                case Mono.Cecil.MethodAttributes.Assembly:
-                case Mono.Cecil.MethodAttributes.FamANDAssem:
+                case 0: // This represents no CodeDOM keyword being specified.
+                case MemberAttributes.Private:
+                case MemberAttributes.Assembly:
+                case MemberAttributes.FamilyAndAssembly:
                     return false;
                 default:
                     return true;
@@ -498,14 +499,15 @@ namespace PublicApiGenerator
 
         static void AddCtorToTypeDeclaration(CodeTypeDeclaration typeDeclaration, MethodDefinition member, AttributeFilter attributeFilter)
         {
-            if (!ShouldIncludeMethod(member.Attributes))
+            var attributes = member.GetMethodAttributes();
+            if (!ShouldIncludeMember(attributes))
                 return;
 
             var method = new CodeConstructor
             {
                 CustomAttributes = CreateCustomAttributes(member, attributeFilter),
                 Name = member.Name,
-                Attributes = member.GetMethodAttributes()
+                Attributes = attributes
             };
             PopulateMethodParameters(member, method.Parameters, attributeFilter);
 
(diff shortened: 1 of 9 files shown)


src/PublicApiGeneratorTests/Event_visibility.cs
```diff
@@ -0,0 +1,130 @@
+﻿using PublicApiGeneratorTests.Examples;
+using System;
+using Xunit;
+
+namespace PublicApiGeneratorTests
+{
+    public class Event_visibility :  ApiGeneratorTestsBase
```
jnm2: This file is a direct copy of `Property_visibility` with a few search and replaces and with the tests of differing accessor visibility removed. (Event accessor visibility can differ in .NET assemblies but not in C#.)

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -721,10 +738,13 @@ static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Pr
             var getterAttributes = member.GetMethod != null ? GetMethodAttributes(member.GetMethod) : 0;
             var setterAttributes = member.SetMethod != null ? GetMethodAttributes(member.SetMethod) : 0;
 
-            if (!HasVisiblePropertyMethod(getterAttributes) && !HasVisiblePropertyMethod(setterAttributes))
+            var hasGet = member.GetMethod is object && ShouldIncludeMember(getterAttributes);
+            var hasSet = member.SetMethod is object && ShouldIncludeMember(setterAttributes);
+
+            if (!(hasGet | hasSet))
```
sungam3r: | or || ?

jnm2: Happy to change to suit your preference. My habit is `|` unless there is a chance of skipping any work. `||` is more to type and introduces a branch.

sungam3r: It does not matter

danielmarbach:
@jnm2 this probably requires a rebase on top of master

jnm2:
Rebased. This one was tricky to untangle. I think I did it right, but shout if something looks off!

src/PublicApiGenerator/CecilEx.cs
```diff
@@ -58,41 +58,33 @@ public static bool IsCompilerGenerated(this IMemberDefinition m)
...
 
-        public static bool HasVisiblePropertyMethod(this MemberAttributes attributes)
-        {
-            var access = attributes & MemberAttributes.AccessMask;
-            return access == MemberAttributes.Public ||
-                   access == MemberAttributes.Family ||
-                   access == MemberAttributes.FamilyOrAssembly;
-        }
-
-        public static MemberAttributes GetPropertyAttributes(MemberAttributes getterAttributes, MemberAttributes setterAttributes)
```
danielmarbach: Thinking out loud. Should this be an extension method on the first param because it is placed in an extension file?

jnm2: Not sure I know the answer, good question.

sungam3r: Technically, there is no concept of an _extension file_ or even an _extension class_. There is an _extension method_. And it can be located in any static class. In addition, even the class itself is not called Cecil**Extensions**.

PR #111, opened 2019-11-02
Add tests showing that 'sealed override' members and event scope modifiers are not supported by CodeDOM

jnm2:
Even when CodeDOM is given e.g. `MemberAttributes.Static` for an event, it ignores it.

CodeDOM incorrectly shows `MemberAttributes.Final | MemberAttributes.Override` as `virtual` rather than `sealed override`.

### DIFF
--- a/src/PublicApiGeneratorTests/Event_modifiers.cs
+++ b/src/PublicApiGeneratorTests/Event_modifiers.cs
@@ -1,0 +1,138 @@
+﻿using PublicApiGeneratorTests.Examples;
+using System;
+using Xunit;
+
+namespace PublicApiGeneratorTests
+{
+    public class Event_modifiers : ApiGeneratorTestsBase
+    {
+        [Fact(Skip = "Not supported by CodeDOM")]
+        [Trait("TODO", "Scope modifiers on events not supported by CodeDOM")]
+        public void Should_output_abstract_modifier()
+        {
+            AssertPublicApi<ClassWithAbstractEvent>(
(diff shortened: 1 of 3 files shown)


src/PublicApiGeneratorTests/Class_events.cs
```diff
@@ -1,39 +0,0 @@
-﻿using System;
-using PublicApiGeneratorTests.Examples;
-using Xunit;
-
-namespace PublicApiGeneratorTests
-{
-    public class Class_events : ApiGeneratorTestsBase
-    {
-        [Fact]
-        public void Should_output_event()
```
sungam3r: Did you mistakenly lose this test when moving to another file?

jnm2: Oops, thanks for catching!

src/PublicApiGeneratorTests/Event_modifiers.cs
```diff
@@ -0,0 +1,138 @@
+﻿using PublicApiGeneratorTests.Examples;
+using System;
+using Xunit;
+
+namespace PublicApiGeneratorTests
+{
+    public class Event_modifiers : ApiGeneratorTestsBase
```
jnm2: This file is a direct copy of Property_modifiers with a few search and replaces.

danielmarbach:
I took your tests and made them green, see https://github.com/ApiApprover/ApiApprover/pull/130

PR #102, opened 2019-10-28
[WIP] C# 8 nullable reference type support 

sungam3r:
Rebased #99 

I did not understand how to do rebase right, the branch was deleted and the previous PR closed, I had to create a new one.

fixes #54

PR #115, opened 2019-11-02
C# 8 nullable reference type support

### LINKED ISSUE #54: Feature request: C# 8 nullable reference type support
It would be great if PublicApiGenerator "understood" NullableAttribute, so could print out the API with suitable `?` syntax.
Trickier: exposing whether an API is null-aware at all.

More info: https://codeblog.jonskeet.uk/2019/02/10/nullableattribute-and-c-8/
Sample code that's way hackier than PublicApiGenerator: https://github.com/nodatime/nodatime/tree/master/build/NullableApiReviewer

### LINKED ISSUE #107: Unnecessary spaces when parameters span multiple lines
Not the end of the world. It looks like this fits with https://github.com/ApiApprover/ApiApprover/issues/46.

All my editors are configured to highlight trailing whitespace:

image

### LINKED ISSUE #108: Generic parameters from declaring type are incorrectly placed on nested types
Input:

```cs
public class Foo<T1, T2>
{
    public class Bar
    {
    }

    public class Bar<T3>
    {
    }
}
```

Expected result in green, actual result in red:

```diff
public class Foo<T1, T2>
{
    public Foo() { }
+   public class Bar
-   public class Bar<T1, T2>
    {
        public Bar() { }
    }
+   public class Bar<T3>
-   public class Bar<T1, T2, T3>
    {
        public Bar() { }
    }
}
```

sungam3r:
C# 8 nullable reference type support
volatile without modreq
fixes #54 
fixes #108
fixes #107
readme note + .editorconfig
unmanaged and notnull constraints

This is final rebased and squashed version of #102

### DIFF
--- a/src/PublicApiGenerator/CSharpAlias.cs
+++ b/src/PublicApiGenerator/CSharpAlias.cs
@@ -1,0 +1,42 @@
+namespace PublicApiGenerator
+{
+    internal static class CSharpAlias
+    {
+        public static string Get(string typeName)
+        {
+            switch (typeName)
+            {
+                case "System.Byte":
+                    return "byte";
+                case "System.SByte":
+                    return "sbyte";
+                case "System.Int16":
+                    return "short";
+                case "System.UInt16":
+                    return "ushort";
+                case "System.Int32":
+                    return "int";
+                case "System.UInt32":
+                    return "uint";
+                case "System.Int64":
+                    return "long";
+                case "System.UInt64":
+                    return "ulong";
+                case "System.Single":
+                    return "float";
+                case "System.Double":
+                    return "double";
+                case "System.Decimal":
+                    return "decimal";
+                case "System.Object":
+                    return "object";
+                case "System.String":
+                    return "string";
+                case "System.Boolean":
+                    return "bool";
+                default:
+                    return typeName;
+            }
+        }
+    }
+}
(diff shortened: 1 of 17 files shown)


src/PublicApiGenerator/CSharpAlias.cs
```diff
@@ -0,0 +1,42 @@
+namespace PublicApiGenerator
+{
+    internal static class CSharpAlias
```
jnm2: I'm a fan of this change. 'Alias' means something specific in C#, so I'd call it `CSharpTypeKeyword` or something.

danielmarbach:
Honestly in other circumstances, I would not have taken this squashed version because it contains unrelated things bundled into one commit. Given that you went through so many hoops of getting this done I hesitated to play out that card and went with pulling it

src/PublicApiGenerator/CSharpAlias.cs
sungam3r: Maybe

sungam3r:
I understand that. I am not happy with such a swelling of the code.

jnm2:
It's a bigger burden for the PR author to make more incremental, standalone changes, but I think it's worth it for everyone involved. Reviewers can follow what's happening so much more easily, and the PR author has to think through the work in a different way which I think pushes them to choose better kinds of changes.

PR #119, opened 2019-11-03
Always consider compiler-recognized internal attributes to be public API

### LINKED ISSUE #117: Compiler-recognized internal attributes are public API
Internal (and even private) attributes are just as visible outside the assembly as inside. I disagree with https://github.com/ApiApprover/ApiApprover/issues/29: you are in reality shipping that annotation as visibly in your .NET binary as if the attribute was public. Either public and internal attributes should be equally visible or they should be equally invisible for the purposes of public API analysis.

However, no matter which way you go on that question, certain well-known attributes affect the compiler behavior regardless of their visibility. They must always be shown. Internal attributes are the only safe way to communicate with the compiler when the attributes are not defined in the target framework. The compiler even generates internal attributes for you when you use C# syntax that depends on these attributes.

The critical ones so that a project can evaluate its nullability API:

https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.allownullattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.disallownullattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.doesnotreturnifattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.maybenullattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.maybenullwhenattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.notnullattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.notnullifnotnullattribute
https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.notnullwhenattribute
➡ Others in `System.Diagnostics.CodeAnalysis` over time

Other attributes that affect compiler or runtime behavior which aren't in all target frameworks and which should be displayed as attributes (e.g. not TupleElementNamesAttribute):

https://docs.microsoft.com/dotnet/api/system.serializableattribute
https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute
https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.callerlinenumberattribute
https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.callerfilepathattribute
https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.callermembernameattribute
https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.referenceassemblyattribute
➡ Others in `System.Runtime.CompilerServices` over time

jnm2:
Fixes #117

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -23,9 +23,7 @@ namespace PublicApiGenerator
 
         public static string GeneratePublicApi(Assembly assembly, Type[] includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[] whitelistedNamespacePrefixes = null, string[] excludeAttributes = null)
         {
-            var attributesToExclude = excludeAttributes == null
-                ? SkipAttributeNames
-                : new HashSet<string>(excludeAttributes.Union(SkipAttributeNames));
+            var attributeFilter = new AttributeFilter(excludeAttributes);
 
             using (var assemblyResolver = new DefaultAssemblyResolver())
             {
@@ -41,7 +39,7 @@ namespace PublicApiGenerator
                 }))
                 {
                     var publicApiForAssembly = CreatePublicApiForAssembly(asm, tr => includeTypes == null || includeTypes.Any(t => t.FullName == tr.FullName && t.Assembly.FullName == tr.Module.Assembly.FullName),
-                        shouldIncludeAssemblyAttributes, whitelistedNamespacePrefixes ?? defaultWhitelistedNamespacePrefixes, attributesToExclude);
+                        shouldIncludeAssemblyAttributes, whitelistedNamespacePrefixes ?? defaultWhitelistedNamespacePrefixes, attributeFilter);
                     return RemoveUnnecessaryWhiteSpace(publicApiForAssembly);
                 }
             }
@@ -60,7 +58,7 @@ namespace PublicApiGenerator
 
         // TODO: Assembly references?
         // TODO: Better handle namespaces - using statements? - requires non-qualified type names
-        static string CreatePublicApiForAssembly(AssemblyDefinition assembly, Func<TypeDefinition, bool> shouldIncludeType, bool shouldIncludeAssemblyAttributes, string[] whitelistedNamespacePrefixes, HashSet<string> excludeAttributes)
+        static string CreatePublicApiForAssembly(AssemblyDefinition assembly, Func<TypeDefinition, bool> shouldIncludeType, bool shouldIncludeAssemblyAttributes, string[] whitelistedNamespacePrefixes, AttributeFilter attributeFilter)
         {
             var publicApiBuilder = new StringBuilder();
             var cgo = new CodeGeneratorOptions
@@ -76,7 +74,7 @@ namespace PublicApiGenerator
                 var compileUnit = new CodeCompileUnit();
                 if (shouldIncludeAssemblyAttributes && assembly.HasCustomAttributes)
                 {
-                    PopulateCustomAttributes(assembly, compileUnit.AssemblyCustomAttributes, excludeAttributes);
+                    PopulateCustomAttributes(assembly, compileUnit.AssemblyCustomAttributes, attributeFilter);
                 }
 
                 var publicTypes = assembly.Modules.SelectMany(m => m.GetTypes())
@@ -94,7 +92,7 @@ namespace PublicApiGenerator
 
                     using (NullableContext.Push(publicType))
                     {
-                        var typeDeclaration = CreateTypeDeclaration(publicType, whitelistedNamespacePrefixes, excludeAttributes);
+                        var typeDeclaration = CreateTypeDeclaration(publicType, whitelistedNamespacePrefixes, attributeFilter);
                         @namespace.Types.Add(typeDeclaration);
                     }
                 }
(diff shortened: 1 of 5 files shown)


src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -471,42 +469,6 @@ static string ConvertAttributeToCode(Func<CodeTypeReference, CodeTypeReference>
...
-            "System.Reflection.AssemblyInformationalVersionAttribute",
-            "System.Reflection.AssemblyProductAttribute",
-            "System.Reflection.AssemblyTitleAttribute",
-            "System.Reflection.AssemblyTrademarkAttribute"
-        };
-
-        static bool ShouldIncludeAttribute(CustomAttribute attribute, HashSet<string> excludeAttributes)
-        {
-            var attributeTypeDefinition = attribute.AttributeType.Resolve();
-            return attributeTypeDefinition != null && !excludeAttributes.Contains(attribute.AttributeType.FullName) && (attributeTypeDefinition.IsPublic || attributeTypeDefinition.FullName == "System.Runtime.CompilerServices.NullableAttribute");
```
jnm2: @sungam3r  ` || attributeTypeDefinition.FullName == "System.Runtime.CompilerServices.NullableAttribute"` seemed out of place. I deleted it and no tests failed.

sungam3r: OK. It is from 4da56b14

src/PublicApiGenerator/AttributeFilter.SkippedAttributeNames.cs
```diff
@@ -0,0 +1,37 @@
+using System.Collections.Generic;
+
+namespace PublicApiGenerator
+{
+    partial class AttributeFilter
+    {
+        private static readonly HashSet<string> SkipAttributeNames = new HashSet<string>
```
sungam3r:
```suggestion
        private static readonly HashSet<string> SkippedAttributeNames = new HashSet<string>
```

src/PublicApiGenerator/AttributeFilter.cs
```diff
@@ -0,0 +1,27 @@
+using Mono.Cecil;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace PublicApiGenerator
+{
+    internal sealed partial class AttributeFilter
+    {
+        private readonly HashSet<string> _excludedAttributes;
```
sungam3r: Maybe use `SkipAttributeNames` directly without one more private field?

src/PublicApiGeneratorTests/Well_known_attributes.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            [DoesNotReturn]
+            public void MethodWithBoolParameter([DoesNotReturnIf(true)] bool a) { }
+        }
+    }
+    // ReSharper restore ClassNeverInstantiated.Global
+    // ReSharper restore UnusedMember.Global
+}
+
+namespace System.Diagnostics.CodeAnalysis
```
sungam3r: :)

sungam3r:
LGTM. Yes, it was worth putting filtering by attributes into a separate class. I had this thought.

src/PublicApiGenerator/ApiGenerator.cs
sungam3r: I did not notice this `NullableAttribute` usage. Probably because the line was too long.

src/PublicApiGenerator/AttributeFilter.cs
jnm2: This is not holding the same set of names that SkipAttributeNames holds, though.

sungam3r: I mean to make SkipAttributeNames instance field and append all excluded attributes to it. SkipAttributeNames and excludedAttributes have the same meaning.

jnm2: Isn't that already what I'm doing? Where would the source of the append be stored?

jnm2: I can rename them `_skippedAttributes` and `DefaultSkippedAttributes` but that sounds like `DefaultSkippedAttributes` is not used if any attributes are specified to exclude.

Or `_excludedAttributes` and `AlwaysExcludedAttributes`?

jnm2:
@danielmarbach This is the last thing required before https://github.com/shouldly/shouldly/pull/581 can become a non-draft PR. When you know of a timeframe for a 10.0.0-beta.3 on the nuget.org community feed, can you let me know?

No rush. I really appreciate the help you've been!

src/PublicApiGenerator/AttributeFilter.cs
sungam3r:
```c# 
internal sealed partial class AttributeFilter
{
     private readonly HashSet<string> SkippedAttributeNames = ...;
     public AttributeFilter(IEnumerable<string> excludedAttributes)
     {
          if (excludedAttributes != null)
                 SkippedAttributeNames  = new HashSet<string>(SkippedAttributeNames.Concat(excludedAttributes));
     }
}

sungam3r: Just use one private instance field instead one static (existed) and one instance (new).

danielmarbach:
Did some minor tweaks, added https://github.com/ApiApprover/ApiApprover/pull/121
Will merge once green

src/PublicApiGenerator/AttributeFilter.cs
jnm2: I don't like this for two reasons:

- The SkippedAttributeNames.cs file is less parallel with the RequiredAttributeNames.cs file
- An unnecessary HashSet is built any time `excludedAttributes` is not null.

Why is having one fewer static fields worth that?

sungam3r: Nothing special. The current version suits me.

PR #122, opened 2019-11-04
Order of keywords

### LINKED ISSUE #80: Order of keywords
We have the following class:

    public static class Utilities

However, the generated API is:

    public class static Utilities

Would it be possible to fix the order of keywords?

sungam3r:
fixes #80


danielmarbach:
This is great timing because I was about to use the same approach for the attributes replacement. I'll take yours in first and then I can readjust my work on top of yours

danielmarbach:
This is great timing because I was about to use the same approach for the attributes replacement. I'll take yours in first and then I can readjust my work on top of yours

sungam3r:
OK.

PR #123, opened 2019-11-04
Add syntax for volatile fields

### LINKED ISSUE #116: C#-like syntax for volatile fields
Nothing special is currently done, so `volatile` on a field shows up as `modreq(System.Runtime.CompilerServices.IsVolatile)` on the field type.

(Reasons that it should be highly visible either way, besides the fact that it's a modreq: If you access a volatile field, the compiler emits a `volatile.` prefix before the stfld/ldfld instructions. It should show up in a dramatic way in your public API if the volatility changes because it's a breaking change. It's a binary behavioral breaking change because libraries compiled against an older version of your library will not use `volatile.` when they should, and it's a compile-time break because errors begin showing up if you take a ref to a suddenly-volatile field.)

sungam3r:
Move all "small"/"helper" methods from ApiGenerator to extensions to keep it clean
fixes #116

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -37,22 +37,14 @@ namespace PublicApiGenerator
                     AssemblyResolver = assemblyResolver
                 }))
                 {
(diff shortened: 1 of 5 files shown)


danielmarbach:
Can you for the next time have one commit that add the volatile and another commit that does the cleanup. That simplifies reviewing a lot. I basically had to search through the file changes to be able to see what is the actual change and what is cleanup

sungam3r:
Sure. Sorry, when I remembered that it was necessary to make the first separate commit, it was already too late. Sometimes you forget yourself in the stream of thoughts.

PR #124, opened 2019-11-06
Nullable Timespan

danielmarbach:
@sungam3r found a small regression while testing with NServiceBus API

```
Expected: ▒▒▒    public readonly TimeSpan? NullableTimespan;\r\n        publ▒▒▒
Actual:   ▒▒▒    public readonly System.TimeSpan? NullableTimespan;\r\n     ▒▒▒
```

### DIFF
--- a/src/PublicApiGenerator/CSharpTypeKeyword.cs
+++ b/src/PublicApiGenerator/CSharpTypeKeyword.cs
@@ -36,6 +36,8 @@ namespace PublicApiGenerator
                     return "bool";
                 case "System.Void":
                     return "void";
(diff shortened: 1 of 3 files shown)


danielmarbach:
Never mind. I think it is better to leave System.TimeSpan since there is no real alias

sungam3r:
I also think so, this is the case when the bug turns out to be a feature.

danielmarbach:
Added a bit more tests to show that https://github.com/ApiApprover/ApiApprover/commit/c9e841f84eb0ad785a533f6b9498cca51ec6b593

PR #125, opened 2019-11-06
fix readonly struct ctor

sungam3r:
@danielmarbach We need beta-4 :)

### DIFF
--- a/src/PublicApiGenerator/CodeNormalizer.cs
+++ b/src/PublicApiGenerator/CodeNormalizer.cs
@@ -51,6 +51,7 @@ namespace PublicApiGenerator
 
             gennedClass = gennedClass.Replace("class " + StaticMarker, "static class ");
             gennedClass = gennedClass.Replace("struct " + ReadonlyMarker, "readonly struct ");
+            gennedClass = gennedClass.Replace(ReadonlyMarker, string.Empty); // remove magic marker from readonly struct ctor
             gennedClass = Regex.Replace(gennedClass, @"\r\n|\n\r|\r|\n", Environment.NewLine);
 
             gennedClass = RemoveUnnecessaryWhiteSpace(gennedClass);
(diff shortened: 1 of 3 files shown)


sungam3r:
image

src/PublicApiGenerator/CodeNormalizer.cs
```diff
@@ -51,6 +51,7 @@ public static string NormalizeGeneratedCode(StringWriter writer)
 
             gennedClass = gennedClass.Replace("class " + StaticMarker, "static class ");
```
danielmarbach: Should we also make sure to remove the static marker?

src/PublicApiGeneratorTests/Struct_readonly.cs
```diff
@@ -12,6 +12,19 @@ public void Should_output()
 @"namespace PublicApiGeneratorTests.Examples
 {
     public readonly struct ReadonlyStruct { }
+}");
+        }
+
+        [Fact]
+        public void Should_output_with_ctor()
```
danielmarbach: Could you also add a similar test for a class with a static constructor?

src/PublicApiGenerator/CodeNormalizer.cs
danielmarbach: and maybe while we are at it also remove `AttributeMarker `

sungam3r: static ctor is not a part of public api, I added tests

sungam3r: > and maybe while we are at it also remove AttributeMarker

Let's remove it only if/when we see such a situation.

src/PublicApiGeneratorTests/Struct_readonly.cs
sungam3r: done

sungam3r:
@danielmarbach Could you please publish beta-4?

danielmarbach:
Certainly not ;)

danielmarbach:
It's up. Sorry for the delay

PR #126, opened 2019-11-07
Add nullability annotations

jnm2:
I think this is the entirety of the public API:

```cs
namespace PublicApiGenerator
{
    public static class ApiGenerator
    {
        public static string GeneratePublicApi(
            Assembly assembly,
            Type[]? includeTypes = null,
            bool shouldIncludeAssemblyAttributes = true,
            string[]? whitelistedNamespacePrefixes = null,
            string[]? excludeAttributes = null);
    }
}
```

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -20,7 +20,7 @@ namespace PublicApiGenerator
     {
         static readonly string[] defaultWhitelistedNamespacePrefixes = new string[0];
 
-        public static string GeneratePublicApi(Assembly assembly, Type[] includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[] whitelistedNamespacePrefixes = null, string[] excludeAttributes = null)
+        public static string GeneratePublicApi(Assembly assembly, Type[]? includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[]? whitelistedNamespacePrefixes = null, string[]? excludeAttributes = null)
         {
             var attributeFilter = new AttributeFilter(excludeAttributes);
 
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -22,6 +22,8 @@ namespace PublicApiGenerator
 
         public static string GeneratePublicApi(Assembly assembly, Type[]? includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[]? whitelistedNamespacePrefixes = null, string[]? excludeAttributes = null)
         {
+            if (assembly is null) throw new ArgumentNullException(nameof(assembly));
+
             var attributeFilter = new AttributeFilter(excludeAttributes);
 
             using (var assemblyResolver = new DefaultAssemblyResolver())
--- a/src/PublicApiGeneratorTests/Interface_methods.cs
+++ b/src/PublicApiGeneratorTests/Interface_methods.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using PublicApiGeneratorTests.Examples;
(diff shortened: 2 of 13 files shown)


src/PublicApiGeneratorTests/Interface_methods.cs
```diff
@@ -93,7 +93,7 @@ public interface IMethodsWithParameters
 
         public interface IMethodsWithDefaultParameters
         {
-            void Method1(int intValue = 42, string stringValue = "hello world", Type typeValue = null);
+            void Method1(int intValue = 42, string stringValue = "hello world", Type typeValue = null!);
```
jnm2: In real code, this should be `Type?`. I didn't want to fuss with changing the tests.

Another way to handle this could be to do `#nullable disable` in each test file for the test types section.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -20,8 +20,10 @@ public static class ApiGenerator
     {
         static readonly string[] defaultWhitelistedNamespacePrefixes = new string[0];
 
-        public static string GeneratePublicApi(Assembly assembly, Type[] includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[] whitelistedNamespacePrefixes = null, string[] excludeAttributes = null)
+        public static string GeneratePublicApi(Assembly assembly, Type[]? includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[]? whitelistedNamespacePrefixes = null, string[]? excludeAttributes = null)
```
sungam3r: if you decide to change the method signature, then let's not just place the `?` modifier on the set of parameters, but make new one class of options and move all the parameters there.

jnm2: This is not a change to the method signature. In an enabled nullability context, no other type is reasonable if you default to null. (I.e. they were already billable, you just couldn't see it before. The real change would be making them non-nullable.)

That being said, I already prefer the options type idea. Do you want to do something like https://github.com/ApiApprover/ApiApprover/issues/117#issuecomment-549068432? I'd prefer to do that 8n a separate PR.

An alternative would be to make PublicApiGenerator instantiable and put the same mutable instance properties on it rather than in a new options type.

sungam3r:
> if you decide to change the method signature

I meant that you just somehow changed this line of code.

> Do you want to do something like #117 (comment)? 

Exactly.

> An alternative would be to make PublicApiGenerator instantiable and put the same mutable instance properties on it rather than in a new options type.

Yes, also a good option.

@danielmarbach I think that gradually correcting the objective problems with the support of various language constructs, we came to the idea of slightly improving the design of the generator itself. From a simple set of static methods, at first, a part of the methods separated and became extension methods, then a filter class appeared. The flat state, transmitted through numerous parameters, began to take shape.

jnm2:
> I meant that you just somehow changed this line of code.

I'm not following why changing that line of code is meaningful. Turning the parameters into properties is either a good idea whether or not the line of code is touched or a bad idea whether or not the line of code is touched :D

sungam3r:
> I'm not following why changing that line of code is meaningful.

Nope. Just one more reason to make options.

danielmarbach:
This is a great step in the right direction. I agree we should go for an options type. If any of you wants to give it a stab I would love to take it. But I would prefer to stay with the static method approach with an options type. For a smooth transition I would recommend to obsolete the current method with a hint to the new method. The old method can then call the new one with the option type. We can then remove the obsoleted method in v11

In the meantime I'll pull this in

danielmarbach:
@jnm2 @sungam3r what do you think about https://github.com/ApiApprover/ApiApprover/pull/129 ?

It is not done yet but I wanted to get your early feedback

sungam3r:
I left my comments about #129 .

PR #127, opened 2019-11-07
Enable debugger source-stepping

jnm2:
A .snupkg file will now appear in the build artifacts. When you do `nuget push Foo.nupkg`, it automatically pushes `Foo.snupkg` as long as it is still in the same folder. More details at https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg#publishing-a-symbol-package.

Make sure that when you push, the NuGet source URL that you're pushing to is a v3 URL (`https://api.nuget.org/v3/index.json`), not a v2 URL, or this will not happen.

To step into ApiApprover code from a consuming project like Shouldly, all that is required is to temporarily turn off [Debug > Options > Enable Just My Code](https://docs.microsoft.com/en-us/visualstudio/debugger/just-my-code) and make sure that the VS2019 default NuGet Symbol Server option is still enabled:

image

### DIFF
--- a/src/PublicApiGenerator/PublicApiGenerator.csproj
+++ b/src/PublicApiGenerator/PublicApiGenerator.csproj
@@ -6,6 +6,14 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\PublicApiGenerator.snk</AssemblyOriginatorKeyFile>
+
+    <!-- Enable debugger source-stepping by embedding source files in the PDB. -->
+    <EmbedAllSources>true</EmbedAllSources>
+
+    <!-- The PDBs go in a .snupkg file to be uploaded to the nuget.org symbol server. -->
+    <!-- https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg -->
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
   <ItemGroup>


danielmarbach:
I have been using https://github.com/dotnet/sourcelink

with https://github.com/dotnet/sourcelink#githubcom-and-github-enterprise

do you want to add the github source provider as well to make this more complete?

jnm2:
@danielmarbach Would you rather use SourceLink instead of embedding the source in the PDBs? (Which are not in the .nupkg, only in the .snupkg?)

I used `<EmbedAllSources>true</EmbedAllSources>` because no network connectivity or VS prompts are needed in order to pull the source. There are fewer things that can go wrong, like how GitHub's raw user content host serves line endings, etc. But SourceLink is easy if that's what you prefer.

danielmarbach:
I'm good either way and I'm OK to support your reasoning about the network connectivity and the VS prompt needed

danielmarbach:
I did some research and talked to a few colleagues. Arguably this is a heated topic as I found out. I'm trying to paraphrase what I learned

Microsoft seemed to have been pushing for a long time for separate symbol packages and the `.snupkg` is just the latest iteration of that. 

The biggest argument from Microsoft perspective on the nuget site seems to be that the package gets larger. That concern has to be largely negated by the portable PDB format.

Bundling the PDB into the package would be simple and straight forward and then you would have the files on the disk when you need it. Putting them in separate packages that then rely on tooling to know how to pull them down in some cases is more complicated, and more error-prone.

I'm now thinking about the embedded option

https://twitter.com/kirillosenkov/status/1007052524946255872?lang=de

danielmarbach:
@jnm2 here we go https://github.com/ApiApprover/ApiApprover/pull/128

jnm2:
@danielmarbach Before the NuGet symbol server was in VS by default, I was on the other side of the argument. I already chalked that you're using portable PDBs
However, as a user of your library, I don't want embedded PDBs because I want to have a clear separation between what I redistribute for runtime and what artifacts I store for debugging.

stakx:
@danielmarbach, it is nice to see a different viewpoint on `.snupkg` packages. They may indeed not be the answer to everything; however, they do offer a reasonably good compromise at the present time. I'd like to share what won me over to using them:

Some people do not *want* debugging symbols; they may interfere with some workflows and tools. If you put them in the main package, or even embed them in the main assembly DLL, people no longer have the option of suppressing or removing debugging symbols. With `.snupkg` and SourceLink, consumers of your package *can* opt out, if they want (albeit in a very coarse-grained way).

When I had to decide whether and how to publish debugging symbols for the moq/moq4 project, I opted to put the PDB directly in the main NuGet package. Because of file size, I switched to the much smaller portable PDBs. It then turned out that some tools (VSTS) couldn't process those yet, as portable PDBs were still quite new; some people had to manually remove the packaged PDBs to avoid failures of their VSTS CI runs.

Granted, the tooling support problem I just described has since been solved, at least for VSTS. But it goes to show that separating concerns, as @jnm2 argued, can generally be beneficial even when not strictly necessary in the majority of all scenarios.

PR #128, opened 2019-11-08
Switching to embedded pdb and source

danielmarbach:
| Scenario                              | Size  |
|---------------------------------------|-------|
| Without PDB and sources               | 52736 |
| With embedded PDB and without sources | 58368 |
| With embedded PDB and sources         | 74240 |

### DIFF
--- a/src/PublicApiGenerator/PublicApiGenerator.csproj
+++ b/src/PublicApiGenerator/PublicApiGenerator.csproj
@@ -6,14 +6,14 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\PublicApiGenerator.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
 
+  <PropertyGroup>
     <!-- Enable debugger source-stepping by embedding source files in the PDB. -->
     <EmbedAllSources>true</EmbedAllSources>
-
-    <!-- The PDBs go in a .snupkg file to be uploaded to the nuget.org symbol server. -->
-    <!-- https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg -->
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>embedded</DebugType>
+    <IncludeSymbols>false</IncludeSymbols>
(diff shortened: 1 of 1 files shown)


danielmarbach:
@jnm2 would love to hear your thoughts

jnm2:
@danielmarbach Oops, I'll copy them here then.

Before the NuGet symbol server was in VS by default, I was on the other side of the argument. I already checked that you're using portable PDBs. My argument was that disk space is so cheap that it is typically a negligible concern, and PDB files in the main nupkg is simply the best for discovery. (It's a really good option and one that you'll see I recommended for NUnit and Shouldly at the time.)

However, as a user of your library, I don't want embedded PDBs because I want to have a clear separation between what I redistribute for runtime and what artifacts I store for debugging.

jnm2:
One drawback of not embedding is that the new SDK csproj format doesn't copy out the PDBs into the build output next to the DLLs unless you paste a workaround such as in
https://github.com/nunit/docs/wiki/Debugger-Source-Stepping.

This is slated to be fixed in this month's .NET Core SDK 3.1.100 release: https://github.com/dotnet/sdk/issues/1458

danielmarbach:
Following your argument wouldn't then the best solution be to use

```
 <PropertyGroup>
    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
  </PropertyGroup>  
```

in combination with embedd sources?

because that puts the PDBs into the nuget package which makes sure they are in the output folder.

Then we wouldn't have to rely on tooling right?

jnm2:
> Following your argument wouldn't then the best solution be to use

Yes, this is what I'm suggesting.

> because that puts the PDBs into the nuget package which makes sure they are in the output folder.

Only if using:

- old-style csproj
- new-style csproj with .NET Core 1.x and 2.x (broken on .NET Core 3.x and on .NET Framework, hopefully fixed this month: dotnet/sdk#1458)
- new-style csproj with pasted workaround (e.g. step 4 of https://github.com/nunit/docs/wiki/Debugger-Source-Stepping)

danielmarbach:
So like this https://github.com/ApiApprover/ApiApprover/pull/131

Sorry I haven't really done this before and I'm ramping up my knowledge here. Thanks for your patience

jnm2:
No problem! Exactly like #131.

danielmarbach:
Closing in favor of https://github.com/ApiApprover/ApiApprover/pull/131

PR #129, opened 2019-11-08
Introduce Options to configure the generator

danielmarbach:


src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -41,14 +41,30 @@ public static string GeneratePublicApi(Assembly assembly, Type[]? includeTypes =
                 {
                     return CreatePublicApiForAssembly(
                         asm,
-                        typeDefinition => includeTypes == null || includeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName),
-                        shouldIncludeAssemblyAttributes,
-                        whitelistedNamespacePrefixes ?? defaultWhitelistedNamespacePrefixes,
+                        typeDefinition => options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName),
```
danielmarbach: For now I decided to step on this level and not float the option type into the underlying method calls

sungam3r: It is OK.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -18,13 +18,13 @@ namespace PublicApiGenerator
 {
     public static class ApiGenerator
     {
-        static readonly string[] defaultWhitelistedNamespacePrefixes = new string[0];
-
-        public static string GeneratePublicApi(Assembly assembly, Type[]? includeTypes = null, bool shouldIncludeAssemblyAttributes = true, string[]? whitelistedNamespacePrefixes = null, string[]? excludeAttributes = null)
+        public static string GeneratePublicApi(Assembly assembly, ApiGeneratorOptions? options = null)
```
sungam3r: Let's make it extension method. It will also be convenient to make such an extension method on a `Type`. In fact the `ApiGeneratorTestsBase` class suffers from the lack of this extension :)

danielmarbach: I'll pull it in with the options type. Happy to accept PRs to further improve this

PR #130, opened 2019-11-08
Event modifier support

### LINKED ISSUE #113: Static/new/virtual/abstract/sealed/override keywords are missing for events
I tried to fix it but it looks like abandoning CodeDOM might be necessary. See https://github.com/ApiApprover/ApiApprover/pull/111.

danielmarbach:
Fixes #113

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -684,7 +684,6 @@ namespace PublicApiGenerator
             if (!(ShouldIncludeMember(addAccessorAttributes) || ShouldIncludeMember(removeAccessorAttributes)))
                 return;
 
-
             var @event = new CodeMemberEvent
             {
                 Name = EventNameBuilder.AugmentEventNameWithEventModifierMarkerTemplate(eventDefinition, eventDefinition.Name, addAccessorAttributes, removeAccessorAttributes),
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -686,7 +686,7 @@ namespace PublicApiGenerator
 
             var @event = new CodeMemberEvent
             {
-                Name = EventNameBuilder.AugmentEventNameWithEventModifierMarkerTemplate(eventDefinition, eventDefinition.Name, addAccessorAttributes, removeAccessorAttributes),
+                Name = EventNameBuilder.AugmentEventNameWithEventModifierMarkerTemplate(eventDefinition, addAccessorAttributes, removeAccessorAttributes),
                 Attributes = CecilEx.CombineAccessorAttributes(addAccessorAttributes, removeAccessorAttributes),
                 CustomAttributes = CreateCustomAttributes(eventDefinition, attributeFilter),
                 Type = eventDefinition.EventType.CreateCodeTypeReference(eventDefinition)
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -684,9 +684,10 @@ namespace PublicApiGenerator
             if (!(ShouldIncludeMember(addAccessorAttributes) || ShouldIncludeMember(removeAccessorAttributes)))
                 return;
 
+
             var @event = new CodeMemberEvent
             {
-                Name = eventDefinition.Name,
+                Name = EventNameBuilder.AugmentEventNameWithEventModifierMarkerTemplate(eventDefinition, eventDefinition.Name, addAccessorAttributes, removeAccessorAttributes),
                 Attributes = CecilEx.CombineAccessorAttributes(addAccessorAttributes, removeAccessorAttributes),
                 CustomAttributes = CreateCustomAttributes(eventDefinition, attributeFilter),
                 Type = eventDefinition.EventType.CreateCodeTypeReference(eventDefinition)
--- a/src/PublicApiGenerator/CodeNormalizer.cs
+++ b/src/PublicApiGenerator/CodeNormalizer.cs
@@ -19,6 +19,8 @@ namespace PublicApiGenerator
         internal const string StaticMarker = "static_C91E2709_C00B-4CAB_8BBC_B2B11DC75E50 ";
         internal const string ReadonlyMarker = "readonly_79D3ED2A_0B60_4C3B_8432_941FE471A38B ";
         internal const string AttributeMarker = "_attribute_292C96C3_C42E_4C07_BEED_73F5DAA0A6DF_";
+        internal const string EventModifierMarkerTemplate = "_{0}_292C96C3C42E4C07BEED73F5DAA0A6DF_";
+        internal const string EventRemovePublicMarker = "removepublic";
 
         public static string NormalizeGeneratedCode(StringWriter writer)
         {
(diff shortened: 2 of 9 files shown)


danielmarbach:
@jnm2 Not exactly a beauty but it works and fixes the 

````
public event 
```

problem on interfaces as well

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -668,9 +668,10 @@ static void AddEventToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Event
             if (!(ShouldIncludeMember(addAccessorAttributes) || ShouldIncludeMember(removeAccessorAttributes)))
                 return;
 
+
```
jnm2: Added on purpose?

danielmarbach: Fixed

danielmarbach:
Green now

src/PublicApiGenerator/CodeNormalizer.cs
```diff
@@ -19,6 +20,8 @@ internal static class CodeNormalizer
         internal const string StaticMarker = "static_C91E2709_C00B-4CAB_8BBC_B2B11DC75E50 ";
         internal const string ReadonlyMarker = "readonly_79D3ED2A_0B60_4C3B_8432_941FE471A38B ";
         internal const string AttributeMarker = "_attribute_292C96C3_C42E_4C07_BEED_73F5DAA0A6DF_";
+        internal const string EventModifierMarkerTemplate = "_{0}_292C96C3C42E4C07BEED73F5DAA0A6DF_";
+        internal const string EventRemovePublicMarker = "removepublic";
```
jnm2: Any reason not to use a GUID to make collisions even less likely?

src/PublicApiGenerator/EventNameBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+
+                return $"{string.Format(CodeNormalizer.EventModifierMarkerTemplate, CodeNormalizer.EventRemovePublicMarker)}{name}";;
+            }
+
+            var addScopeAttributes = (addAccessorAttributes & MemberAttributes.ScopeMask);
+            var removeScopeAttributes = (removeAccessorAttributes & MemberAttributes.ScopeMask);
+
+            if (addScopeAttributes == MemberAttributes.Static && removeScopeAttributes == MemberAttributes.Static)
+            {
+                return $"{string.Format(CodeNormalizer.EventModifierMarkerTemplate, "static")}{name}";
```
jnm2: I think people recommend to use concatenation whenever interpolation is only used to insert things at the very beginning or end of the string.

src/PublicApiGenerator/EventNameBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            if (addScopeAttributes == (MemberAttributes.Final | MemberAttributes.Override) && removeScopeAttributes == (MemberAttributes.Final | MemberAttributes.Override))
+            {
+                return $"{string.Format(CodeNormalizer.EventModifierMarkerTemplate, "sealed override")}{name}";
+            }
+
+            if (addScopeAttributes == MemberAttributes.Abstract && removeScopeAttributes == MemberAttributes.Abstract)
+            {
+                return $"{string.Format(CodeNormalizer.EventModifierMarkerTemplate, "abstract")}{name}";
+            }
```
jnm2: Instead of repeating the check in every `if` statement, you could skip all these statements unless `addScopeAttributes == removeScopeAttributes`. Then the `if` statements contained by that check could be converted to a switch on `addScopeAttributes`.

src/PublicApiGenerator/EventNameBuilder.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            if (addScopeAttributes == (MemberAttributes.Final | MemberAttributes.Override) && removeScopeAttributes == (MemberAttributes.Final | MemberAttributes.Override))
+            {
+                return $"{string.Format(CodeNormalizer.EventModifierMarkerTemplate, "sealed override")}{name}";
+            }
+
+            if (addScopeAttributes == MemberAttributes.Abstract && removeScopeAttributes == MemberAttributes.Abstract)
+            {
+                return $"{string.Format(CodeNormalizer.EventModifierMarkerTemplate, "abstract")}{name}";
```
jnm2: It looks like it's not possible for the method to combine `new` and `abstract`, but it should be able to.

src/PublicApiGenerator/CodeNormalizer.cs
danielmarbach: The marker becomes part of the `_{0}_GUID` signature which is matched. So it is already unique enough

src/PublicApiGenerator/EventNameBuilder.cs
jnm2: Also `static new`

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -670,7 +670,7 @@ static void AddEventToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Event
 
             var @event = new CodeMemberEvent
             {
-                Name = eventDefinition.Name,
+                Name = EventNameBuilder.AugmentEventNameWithEventModifierMarkerTemplate(eventDefinition, eventDefinition.Name, addAccessorAttributes, removeAccessorAttributes),
```
sungam3r: `eventDefinition.Name` argument seems redundant

sungam3r:
Replacement magic back and forth :)

src/PublicApiGenerator/EventNameBuilder.cs
danielmarbach: Dude you sent me down a rabbit hole 👯‍♂

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: Fixed

src/PublicApiGenerator/EventNameBuilder.cs
danielmarbach: Done

src/PublicApiGenerator/EventNameBuilder.cs
danielmarbach: Done

danielmarbach:
OMG. I feel dirty now, really dirty

src/PublicApiGenerator/EventNameBuilder.cs
danielmarbach: STOP THIS MADNESS ;). If there is no test it doesn't exist 💃

danielmarbach: Done

src/PublicApiGenerator/EventNameBuilder.cs
```diff
@@ -0,0 +1,66 @@
...
+                {
+                    break;
+                }
+                baseType = typeDef?.BaseType;
+            }
+
+            var addScopeAttributes = addAccessorAttributes & MemberAttributes.ScopeMask;
+            switch (addScopeAttributes)
+            {
+                case MemberAttributes.Static when addScopeAttributes == MemberAttributes.Static && !isNew.HasValue:
```
sungam3r: Why one more `addScopeAttributes == MemberAttributes.Static` check here and below?

jnm2:
@danielmarbach Sorry, I forgot about `abstract override`.

src/PublicApiGenerator/EventNameBuilder.cs
jnm2: In each `when`, you start with a redundant check. It's impossible for it to be otherwise since we've reached the `when` after this `case`.

jnm2: (If you meant `removeScopeAttributes`, I'd say that check should be `addScopeAttributes == removeScopeAttributes` and should be before the switch.)

danielmarbach: I check this in the entry. Yeah noticed now as well. removed the redundant cases

jnm2:
> @danielmarbach Sorry, I forgot about abstract override.

Wouldn't it be better to have a separate check for each keyword in order and append them as you go, so that you don't have to special-case every possible combination?

danielmarbach:
Maybe but I'm done here https://github.com/ApiApprover/ApiApprover/pull/133

PR #136, opened 2019-11-10
Propagate DefaultMemberAttribute to IndexerNameAttribute

### LINKED ISSUE #106: Named indexers are shown as parameterless properties
Compiled code:

```c#
public class Foo
{
    [IndexerName("Bar")]
    public int this[int x] => x;
}
```

Actual result:

```cs
public class Foo
{
    public Foo() { }
    public int Bar { get; }
}
```

Expected result:

```c#
public class Foo
{
    [IndexerName("Bar")]
    public int this[int x] { get; }
}
```

Alternative expected result (if the lower-level representation is your preference):

```c#
[DefaultMember("Bar")]
public class Foo
{
    public int this[int x] { get; }
}
```

danielmarbach:
Closes #106

I think soon I will need to cleanup the dependencies between the names and the code normalizer a bit but for now the name uses the markers defined on the normalizer

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -685,16 +685,18 @@ namespace PublicApiGenerator
             };
 
             // DefaultMemberAttribute on type gets propagated to IndexerNameAttribute
-            var defaultMemberAttribute = typeDeclarationInfo.CustomAttributes.SingleOrDefault(x =>
-                x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute");
-            var defaultMemberAttributeArgument = defaultMemberAttribute?.ConstructorArguments.SingleOrDefault();
-            var value = defaultMemberAttributeArgument?.Value as string;
-            if (!string.IsNullOrEmpty(value) && propertyName != "Item")
+            var defaultMemberAttributeValue = typeDeclarationInfo.CustomAttributes.SingleOrDefault(x =>
+                    x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute")
+                ?.ConstructorArguments.Select(x => x.Value).OfType<string>().SingleOrDefault();
+            if (!string.IsNullOrEmpty(defaultMemberAttributeValue) && propertyName != "Item")
             {
                 property.Name = "Item";
-                var codeAttributeDeclaration = new CodeAttributeDeclaration(AttributeNameBuilder.Get("System.Runtime.CompilerServices.IndexerNameAttribute"));
-                codeAttributeDeclaration.Arguments.Add(new CodeAttributeArgument(new CodePrimitiveExpression(value)));
-                property.CustomAttributes.Add(codeAttributeDeclaration);
+                property.CustomAttributes.Add(
+                    new CodeAttributeDeclaration(
+                        AttributeNameBuilder.Get("System.Runtime.CompilerServices.IndexerNameAttribute"))
+                    {
+                        Arguments = {new CodeAttributeArgument(new CodePrimitiveExpression(defaultMemberAttributeValue))}
+                    });
             }
 
             // Here's a nice hack, because hey, guess what, the CodeDOM doesn't support
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -180,6 +180,7 @@ namespace PublicApiGenerator
         }
 
         static void AddMemberToTypeDeclaration(CodeTypeDeclaration typeDeclaration,
+            IMemberDefinition typeDeclarationInfo,
             IMemberDefinition memberInfo,
             AttributeFilter attributeFilter)
         {
@@ -194,7 +195,7 @@ namespace PublicApiGenerator
                 }
                 else if (memberInfo is PropertyDefinition propertyDefinition)
                 {
-                    AddPropertyToTypeDeclaration(typeDeclaration, propertyDefinition, attributeFilter);
+                    AddPropertyToTypeDeclaration(typeDeclaration, typeDeclarationInfo, propertyDefinition, attributeFilter);
                 }
                 else if (memberInfo is EventDefinition eventDefinition)
                 {
@@ -293,14 +294,14 @@ namespace PublicApiGenerator
                 declaration.BaseTypes.Add(@interface.InterfaceType.CreateCodeTypeReference(@interface));
 
             foreach (var memberInfo in publicType.GetMembers().Where(memberDefinition => ShouldIncludeMember(memberDefinition, whitelistedNamespacePrefixes)).OrderBy(m => m.Name, StringComparer.Ordinal))
-                AddMemberToTypeDeclaration(declaration, memberInfo, attributeFilter);
+                AddMemberToTypeDeclaration(declaration, publicType, memberInfo, attributeFilter);
 
             // Fields should be in defined order for an enum
             var fields = !publicType.IsEnum
                 ? publicType.Fields.OrderBy(f => f.Name, StringComparer.Ordinal)
                 : (IEnumerable<FieldDefinition>)publicType.Fields;
             foreach (var field in fields)
-                AddMemberToTypeDeclaration(declaration, field, attributeFilter);
+                AddMemberToTypeDeclaration(declaration, publicType, field, attributeFilter);
 
             foreach (var nestedType in publicType.NestedTypes.Where(ShouldIncludeType).OrderBy(t => t.FullName, StringComparer.Ordinal))
             {
@@ -445,10 +446,7 @@ namespace PublicApiGenerator
         static CodeAttributeDeclaration GenerateCodeAttributeDeclaration(Func<CodeTypeReference, CodeTypeReference> codeTypeModifier, CustomAttribute customAttribute)
         {
             var attribute = new CodeAttributeDeclaration(codeTypeModifier(customAttribute.AttributeType.CreateCodeTypeReference(mode: NullableMode.Disable)));
-            if (attribute.Name != "System.ParamArrayAttribute")
-            {
-                attribute.Name = $"{attribute.Name}{CodeNormalizer.AttributeMarker}";
-            }
+            attribute.Name = AttributeNameBuilder.Get(attribute.Name);
             foreach (var arg in customAttribute.ConstructorArguments)
             {
                 attribute.Arguments.Add(new CodeAttributeArgument(CreateInitialiserExpression(arg)));
@@ -658,7 +656,7 @@ namespace PublicApiGenerator
             return parameter.ParameterType.IsValueType ? "default" : "null";
         }
 
-        static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, PropertyDefinition member, AttributeFilter attributeFilter)
+        static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, IMemberDefinition typeDeclarationInfo, PropertyDefinition member, AttributeFilter attributeFilter)
         {
             var getterAttributes = member.GetMethod?.GetMethodAttributes() ?? 0;
             var setterAttributes = member.SetMethod?.GetMethodAttributes() ?? 0;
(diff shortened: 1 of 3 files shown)


danielmarbach:
@jnm2 What do you think?

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -675,16 +673,30 @@ static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Pr
...
+                Name = propertyName,
                 Type = propertyType,
                 Attributes = propertyAttributes,
                 CustomAttributes = CreateCustomAttributes(member, attributeFilter),
                 HasGet = hasGet,
                 HasSet = hasSet
             };
 
+            // DefaultMemberAttribute on type gets propagated to IndexerNameAttribute
+            var defaultMemberAttribute = typeDeclarationInfo.CustomAttributes.SingleOrDefault(x =>
```
danielmarbach: Was thinking of pulling this somewhere else but then I figured many of these generator methods have custom logic and this one doesn't make the already crappy situation much worse

src/PublicApiGenerator/AttributeNameBuilder.cs
```diff
@@ -0,0 +1,10 @@
+namespace PublicApiGenerator
+{
+    public static class AttributeNameBuilder
+    {
+        public static string Get(string name)
+        {
+            return name != "System.ParamArrayAttribute" ? $"{name}{CodeNormalizer.AttributeMarker}" : name;
```
sungam3r:
```suggestion
            return name == "System.ParamArrayAttribute" ? name : $"{name}{CodeNormalizer.AttributeMarker}";
```

src/PublicApiGeneratorTests/Indexer_properties.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public class ClassWithIndexer
+        {
+            public int this[int x] => x;
+        }
+
+        public class ClassWithNamedIndexer
+        {
+            [IndexerName("Bar")]
+            public int this[int x] => x;
```
sungam3r: It would be nice to add `{ get; set; }` indexer too.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -675,16 +673,30 @@ static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Pr
...
+                x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute");
+            var defaultMemberAttributeArgument = defaultMemberAttribute?.ConstructorArguments.SingleOrDefault();
+            var value = defaultMemberAttributeArgument?.Value as string;
+            if (!string.IsNullOrEmpty(value) && propertyName != "Item")
+            {
+                property.Name = "Item";
```
sungam3r: What would happen without this line?

jnm2: I'm guessing this is the line that makes CodeDOM stop rendering it as a non-indexer property.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -675,16 +673,30 @@ static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Pr
...
+                var codeAttributeDeclaration = new CodeAttributeDeclaration(AttributeNameBuilder.Get("System.Runtime.CompilerServices.IndexerNameAttribute"));
+                codeAttributeDeclaration.Arguments.Add(new CodeAttributeArgument(new CodePrimitiveExpression(value)));
+                property.CustomAttributes.Add(codeAttributeDeclaration);
```
jnm2: (option)

```cs
property.CustomAttributes.Add(
    new CodeAttributeDeclaration(AttributeNameBuilder.Get("System.Runtime.CompilerServices.IndexerNameAttribute"))
    {
        Arguments = { new CodeAttributeArgument(new CodePrimitiveExpression(value)) }
    });
```

src/PublicApiGenerator/ApiGenerator.cs
jnm2: `SingleOrDefault` makes a hard assumption that no one will ever declare and use a constructor with two parameters on an attribute with the name `System.Reflection.DefaultMemberAttribute`, which is probably true.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -675,16 +673,30 @@ static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Pr
...
                 Attributes = propertyAttributes,
                 CustomAttributes = CreateCustomAttributes(member, attributeFilter),
                 HasGet = hasGet,
                 HasSet = hasSet
             };
 
+            // DefaultMemberAttribute on type gets propagated to IndexerNameAttribute
+            var defaultMemberAttribute = typeDeclarationInfo.CustomAttributes.SingleOrDefault(x =>
+                x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute");
+            var defaultMemberAttributeArgument = defaultMemberAttribute?.ConstructorArguments.SingleOrDefault();
```
jnm2: This `SingleOrDefault` makes a hard assumption that no one will ever declare and use a constructor with two parameters on an attribute with the name `System.Reflection.DefaultMemberAttribute`, which is probably true.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -675,16 +673,30 @@ static void AddPropertyToTypeDeclaration(CodeTypeDeclaration typeDeclaration, Pr
...
+            var value = defaultMemberAttributeArgument?.Value as string;
+            if (!string.IsNullOrEmpty(value) && propertyName != "Item")
```
jnm2: I think this should be `if (value == propertyName)` in case there are two of these things. (VB allows you to declare any number of named indexers, but only one can be the default indexer.)

src/PublicApiGenerator/ApiGenerator.cs
jnm2: (option)

```cs
var defaultMember = typeDeclarationInfo.CustomAttributes
    .SingleOrDefault(x => x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute"
    ?.ConstructorArguments.SingleOrDefault()
    ?.Value as string;
```

src/PublicApiGenerator/ApiGenerator.cs
jnm2: I think this should be `if (value == propertyName)` in case there are two of these things in the same class with different names. (VB allows you to declare any number of named indexers, but only one can be the default indexer.)

jnm2: So the question is, how do you want this to render?

```vb
Public Class C
    Public ReadOnly Default Property Foo(i As Integer) As Integer
        Get
            Return i
        End Get
    End Property
    
    Public ReadOnly Property Bar(i As Integer) As Integer
        Get
            Return i
        End Get
    End Property
End Class
```

Right now because `propertyName` is not compared to the default member name, you render them both with `[IndexerName("Foo")]` which is misleading:

```
public class C
{
    [IndexerName("Foo")]
    public int this[int i] { get; }

    [IndexerName("Foo")]
    public int this[int i] { get; }
}
```



[(sharplab)](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AbEA3YaAuIAlhgD4AKArsBkWAAQDCGAhgM5tMCwAUPf/So069AEowWAEwDyAOwwBPegBEYAMxaUM+QVAgAHGFHxKAYhAgAKIvQCCnAJKz8MAOZGAlHcfO3R3gKB9ADiMPgBQZHi+JRQsvREEZEAorKSIWFJ9Knp5HqGxgpZWUK0DOJScoq6BkYm9ABCLFDW3vROLu5QXvbtvl1ZgaHhfJFB0bHxiaMpaRkjgTk1BSa8S8zsbLxAA==)

jnm2: So the question is, how do you want this to render?

```vb
Public Class C
    Public ReadOnly Default Property Foo(i As Integer) As Integer
        Get
            Return i
        End Get
    End Property
    
    Public ReadOnly Property Bar(i As Integer) As Integer
        Get
            Return i
        End Get
    End Property
End Class
```

Right now because `propertyName` is not compared to the default member name, you render them both as default indexers which is slightly misleading, but I don't know what the alternative would be:

```
public class C
{
    [IndexerName("Foo")]
    public int this[int i] { get; }

    [IndexerName("Bar")]
    public int this[int i] { get; }
}
```

[(sharplab)](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AbEA3YaAuIAlhgD4AKArsBkWAAQDCGAhgM5tMCwAUPf/So069AEowWAEwDyAOwwBPegBEYAMxaUM+QVAgAHGFHxKAYhAgAKIvQCCnAJKz8MAOZGAlHcfO3R3gKB9ADiMPgBQZHi+JRQsvREEZEAorKSIWFJ9Knp5HqGxgpZWUK0DOJScoq6BkYm9ABCLFDW3vROLu5QXvbtvl1ZgaHhfJFB0bHxiaMpaRkjgTk1BSa8S8zsbLxAA==)

jnm2: Maybe this would be the best. The next question is whether it's worth any effort. Will anyone that uses VB use a tool that shows them their API in C#-like syntax? And will they ever use a non-default named indexer?

```cs
public class C
{
    [IndexerName("Foo")]
    public int this[int i] { get; }

    public int Bar[int i] { get; } 
}
```

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: Correct. It requires Item

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: So far I haven't heard anything from VB users. And if it becomes a problem we can fix it in another version. Yagni for now

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: Unless someone would redeclare the property in that namespace yes.

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: I honestly don't care both are good for me.

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: I find this train wreck harder to read but I can live with your option as well

src/PublicApiGenerator/AttributeNameBuilder.cs
danielmarbach: I prefer my variant but I'm fine either way. Why do you prefer your suggestion?

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: I think rendering both as default is ok for now for such an edge case I don't want to increase the complexity because I would probably need to fallback to fancy string replacement magic again. I'd rather save my energy for the method and property declaration fixes. And maybe we can there beat this edge case because we probably require then anyway fancy property naming and replacement trickery like what we have done with events.

src/PublicApiGenerator/AttributeNameBuilder.cs
sungam3r: I assume that code is easier to read when there are fewer negative conditions. Then it is not necessary to mentally invert the condition.

src/PublicApiGenerator/ApiGenerator.cs
danielmarbach: Now I remember again, 

image

```
  ApiGenerator.cs(691, 17): [CS0023] Operator '?' cannot be applied to operand of type 'CustomAttributeArgument'
```

danielmarbach:
```
            var defaultMemberAttributeValue = typeDeclarationInfo.CustomAttributes.SingleOrDefault(x =>
                x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute")
                ?.ConstructorArguments.Select(x => x.Value).SingleOrDefault() as string;
```

danielmarbach: or even better

```
            var defaultMemberAttributeValue = typeDeclarationInfo.CustomAttributes.SingleOrDefault(x =>
                    x.AttributeType.FullName == "System.Reflection.DefaultMemberAttribute")
                ?.ConstructorArguments.Select(x => x.Value).OfType<string>()
                .SingleOrDefault();
```

PR #138, opened 2019-11-12
Fix member ordering

adamralph:


src/PublicApiGenerator/CodeTypeDeclarationExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            foreach (var key in original.UserData.Keys)
+            {
+                sorted.UserData[key] = original.UserData[key];
+            }
+
+            var emptyParamList = new List<CodeParameterDeclarationExpression>();
+
+            var sortedMembers = original.Members.OfType<CodeTypeMember>()
+                .OrderBy(m => m.GetType().Name, StringComparer.Ordinal)
```
sungam3r: Properties go ahead of methods?

sungam3r:
Interesting attention to detail 👍

src/PublicApiGenerator/CodeTypeDeclarationExtensions.cs
adamralph: Yes, this appears to have no effect on the order in comparison to what is currently in `master`.

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -312,7 +312,7 @@ static CodeTypeDeclaration CreateTypeDeclaration(TypeDefinition publicType, stri
                 }
             }
 
-            return declaration;
+            return declaration.Sort();
         }
 
         static CodeTypeDeclaration CreateDelegateDeclaration(TypeDefinition publicType, AttributeFilter attributeFilter)
```
danielmarbach: `CreateDelegateDeclaration` creates a special type of `CodeTypeDeclaration` which is of type `CodeTypeDelegate`. I think it would make sense for consistency reason to also sort parameters on delegate declarations. What do you think?

```
                var sorted= new CodeTypeDelegate(@delegate.Name)
                {
                    Attributes = @delegate.Attributes,
                    CustomAttributes = @delegate.CustomAttributes,
                    ReturnType = @delegate.ReturnType,
                };
                sorted.Parameters.AddRange(@delegate.Parameters);
```

sungam3r: OK

adamralph: I'm not sure I understand. I'm not sure there's anything to sort in a delegate. Does your code snippet actually change anything?

danielmarbach: @adamralph your code also sorts by the parameters of a method right? See https://github.com/ApiApprover/ApiApprover/pull/138/files#diff-9cab084db4a89e617009fb6111b48f7eR50-R53

But I guess you are implicitly saying that even though a delegate has parameters it doesn't affect any sorting because it is sort of stand alone?

Or would having a delegate as part of a declaration change anything? And that is why I'm trying to explore if delegate type declarations should also be taken into account in the sorting.

adamralph: @danielmarbach a delegate doesn't have members, i..e this is a delegate declaration _in it's entirety_:

```c#
public delegate void Foo();
```

So there is nothing to sort within a `CodeTypeDeclaration` (if undertsand the type correctly) which represents a delegate.

We could sort delegates _within a namespace_ by their type parameters and parameters, similar to how this PR is doing for members _within a type_, but I think that would be best handled elsewhere, because it would form part of how all types (classes, structs, interfaces, delegates, enums, etc.) are sorted within a namespace. I'd be happy to investigate that as well, but I think it should be a seperate PR.

adamralph:
@danielmarbach I realised that having static and instance members jumbled together isn't great either, so I pushed fixup commits to alter the tests and implementation to ensure statics come after instance members. I guess my (slight) preference may be to have statics _before_ instance members, but we'd have to alter some more existing tests to make that change.

danielmarbach:
Honestly I prefer consistency over any member ordering preferences. As long as we have a stable way to sort things we are good and I would try to avoid to go to far into any directions of "I like it this way" because then someone else will come and say "but this other way is much better".

PR #141, opened 2019-11-13
make second param of GeneratePublicApi truly optional, and align extension methods

### LINKED ISSUE #139: 10.0.0 changes make it impossible to call GeneratePublicApi with just an assembly argument
```c#
ApiGenerator.GeneratePublicApi(typeof(Foo).Assembly));
```

> The call is ambiguous between the following methods or properties: 'ApiGenerator.GeneratePublicApi(Assembly, ApiGeneratorOptions?)' and 'ApiGenerator.GeneratePublicApi(Assembly, Type[]?, bool, string[]?, string[]?)'

This worked in 9.x.

adamralph:
fixes #139

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -76,14 +76,14 @@ namespace PublicApiGenerator
         /// <param name="type">The type to generate an API from.</param>
         /// <param name="options">The options to control the API output.</param>
         /// <returns>The API output.</returns>
-        public static string GeneratePublicApi(this Type type, ApiGeneratorOptions? options = null)
+        public static string GeneratePublicApi(this Type type, ApiGeneratorOptions? options)
         {
             (options ??= new ApiGeneratorOptions()).IncludeTypes = new Type[] { type };
             return type.Assembly.GeneratePublicApi(options);
         }
 
         [Obsolete("Use `GeneratePublicApi(this Assembly assembly, ApiGeneratorOptions? options = null)` instead. Will be removed in the next major.")]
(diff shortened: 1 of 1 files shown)


sungam3r:
When obsolete overload will be removed it will be possible to return `= null` in the first extension method.

adamralph:
@sungam3r I don't understand what you mean. Are you saying there is a problem with this change?

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -76,14 +76,14 @@ public static string GeneratePublicApi(this Type[] types, ApiGeneratorOptions? o
...
         /// <param name="options">The options to control the API output.</param>
         /// <returns>The API output.</returns>
-        public static string GeneratePublicApi(this Type type, ApiGeneratorOptions? options = null)
+        public static string GeneratePublicApi(this Type type, ApiGeneratorOptions? options)
         {
             (options ??= new ApiGeneratorOptions()).IncludeTypes = new Type[] { type };
             return type.Assembly.GeneratePublicApi(options);
         }
 
         [Obsolete("Use `GeneratePublicApi(this Assembly assembly, ApiGeneratorOptions? options = null)` instead. Will be removed in the next major.")]
```
danielmarbach: Is this message still clear or should we remove the `= null`?

sungam3r:
No problems.
image
I talk about that we can return `= null` for this method in next major version when obsolete method will be removed.

src/PublicApiGenerator/ApiGenerator.cs
sungam3r: To be consistent we should remove `= null`.

adamralph:
@sungam3r thanks, I didn't realise that that method is obsolete. In that case, I don't think this PR is required. Consumers should switch to the other method, and in doing so, can switch extension method syntax which will make the second parameter optional.

sungam3r:
The bug turned into a feature :). So we "motivate" the client to use the new extension method.

PR #143, opened 2019-11-14
Fix virtual event

sungam3r:
Before:
```c#
   public abstract class AbstractClassWithEvent : PublicApiGeneratorTests.Examples.InterfaceWithEvent
    {
        protected AbstractClassWithEvent() { }
        public virtual event System.EventHandler Event;
    }
```
After:
```c#
   public abstract class AbstractClassWithEvent : PublicApiGeneratorTests.Examples.InterfaceWithEvent
    {
        protected AbstractClassWithEvent() { }
        public event System.EventHandler Event;
    }
```

In a separate commits I added several tests. I would like to ask whether the default constructor (protected) is correctly displayed by the abstract class? It seems to me that is right, but I doubt it.


danielmarbach:
Split into #144, #145 and #146

Issue #112, opened 2019-11-02
'Sealed override' is shown as 'virtual'


jnm2:
I tried to fix it but it looks like abandoning CodeDOM might be necessary. See https://github.com/ApiApprover/ApiApprover/pull/111.

sungam3r:
> abandoning CodeDOM might be necessary

Or forking it.

danielmarbach:
Once the event modifier stuff is in we could technically do the same trickery with methods

danielmarbach:
I'm not saying I like it but so far Code Dom has proven itself fairly flexible :)

sungam3r:
Of course, with CodeDom and replacement magic you can achieve much more than just with CodeDom.

danielmarbach:
Added https://github.com/PublicApiGenerator/PublicApiGenerator/pull/152 for the method part. I have to finish 6 tests that came from the event modifier changes but other than that it seems that most cases are now addressed

danielmarbach:
PR is ready to go. In case nobody chimes in with thoughts I will merge it tomorrow evening or so

danielmarbach:
Added https://github.com/PublicApiGenerator/PublicApiGenerator/pull/155 for the properties part

danielmarbach:
I think this is reasonably addressed. Pretty sure there is an edge case that isn't covered yet but we can deal with those when the need arises.

danielmarbach:
Do you want to give beta 8 a stab? There is a known issue with a bit too much whitespace with the new modifer that I already sent a PR for. Other than that I hope this is it for 10.0.0

PR #152, opened 2019-11-19
Support for more method modifiers

danielmarbach:
Support for more method scenarios for #112

### DIFF
--- a/src/PublicApiGenerator/CodeNormalizer.cs
+++ b/src/PublicApiGenerator/CodeNormalizer.cs
@@ -24,6 +24,11 @@ namespace PublicApiGenerator
         internal const string EventRemovePublicMarker = "removepublic";
         internal const string MethodModifierMarkerTemplate = "_{0}_3C0D97CD952D40AA8B6E1ECB98FFC79F_";
 
+        public static string NormalizeMethodName(string methodName)
+        {
+            return Regex.Replace(methodName, @"(_(.*)_3C0D97CD952D40AA8B6E1ECB98FFC79F_)?", string.Empty);
+        }
+
         public static string NormalizeGeneratedCode(StringWriter writer)
         {
             var gennedClass = writer.ToString();
--- a/src/PublicApiGenerator/CodeNormalizer.cs
+++ b/src/PublicApiGenerator/CodeNormalizer.cs
@@ -22,6 +22,7 @@ namespace PublicApiGenerator
         internal const string AttributeMarker = "_attribute_292C96C3_C42E_4C07_BEED_73F5DAA0A6DF_";
         internal const string EventModifierMarkerTemplate = "_{0}_292C96C3C42E4C07BEED73F5DAA0A6DF_";
         internal const string EventRemovePublicMarker = "removepublic";
+        internal const string MethodModifierMarkerTemplate = "_{0}_3C0D97CD952D40AA8B6E1ECB98FFC79F_";
 
         public static string NormalizeGeneratedCode(StringWriter writer)
         {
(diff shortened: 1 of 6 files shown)


src/PublicApiGenerator/MethodNameBuilder.cs
```diff
@@ -0,0 +1,53 @@
...
+                // too simple, improve
+                isNew = typeDef?.Methods.Any(e => e.Name.Equals(methodDefinition.Name, StringComparison.Ordinal) && e.Parameters.Count == methodDefinition.Parameters.Count);
+                if (isNew is true)
+                {
+                    break;
+                }
+                baseType = typeDef?.BaseType;
+            }
+
+            return (attributes & MemberAttributes.ScopeMask, isNew, methodDefinition.IsVirtual,
```
sungam3r: it looks like it was copied from another file. Is it worth to extract in a method?

src/PublicApiGenerator/CodeNormalizer.cs
```diff
@@ -83,6 +90,15 @@ static string EventModifierMatcher(Match group)
                 .Replace($"{visibility} event ", replacementBuilder.ToString());
         }
 
+        static string MethodModifierMatcher(Match group)
```
jnm2:
```suggestion
        static string MethodModifierMatcher(Match match)
```

src/PublicApiGenerator/MethodNameBuilder.cs
danielmarbach: I'm not a huge fan of generalizing too early. I'd wait and see how the property cases look like and then decide

src/PublicApiGenerator/MethodNameBuilder.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        public static string AugmentMethodNameWithMethodModifierMarkerTemplate(MethodDefinition methodDefinition,
+            MemberAttributes attributes)
+        {
+            var name = CSharpOperatorKeyword.Get(methodDefinition.Name);
+
+            if (methodDefinition.DeclaringType.IsInterface)
+            {
+                return name;
+            }
```
jnm2: ℹ This is going to have to change in the future to support the many modifiers now allowed on interface members in C#.
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#default-interface-methods

src/PublicApiGenerator/MethodNameBuilder.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            bool? isNew = null;
+            var baseType = methodDefinition.DeclaringType.BaseType;
+            while (baseType != null)
+            {
+                var typeDef = baseType as TypeDefinition;
+                // too simple, improve
+                isNew = typeDef?.Methods.Any(e => e.Name.Equals(methodDefinition.Name, StringComparison.Ordinal) && e.Parameters.Count == methodDefinition.Parameters.Count);
```
jnm2: `e.Name.Equals(methodDefinition.Name, StringComparison.Ordinal)` does the same comparison as `e.Name == methodDefinition.Name`

src/PublicApiGenerator/MethodNameBuilder.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (methodDefinition.DeclaringType.IsInterface)
+            {
+                return name;
+            }
+
+            bool? isNew = null;
+            var baseType = methodDefinition.DeclaringType.BaseType;
+            while (baseType != null)
+            {
+                var typeDef = baseType as TypeDefinition;
```
jnm2: If desired:

```cs
            while (baseType is TypeDefinition typeDef)
            {
```

src/PublicApiGeneratorTests/Method_modifiers.cs
```diff
@@ -20,15 +48,57 @@ public class ClassWithStaticMethod
...
+        [Fact]
+        public void Should_output_static_new_modifier()
+        {
+            AssertPublicApi<ClassWithStaticNewMethod>(
+                @"namespace PublicApiGeneratorTests.Examples
+{
+    public class ClassWithStaticNewMethod : PublicApiGeneratorTests.Examples.ClassWithStaticMethod
+    {
+        public ClassWithStaticNewMethod() { }
+        public new static void DoSomething() { }
```
jnm2: I was thinking this was typically `static new` but I'm not finding a canonical list, so...?

danielmarbach: Honestly I used ReSharper suggestion but will double check

src/PublicApiGenerator/MethodNameBuilder.cs
danielmarbach: Don't get me started... 😂
Yeah but we can focus on that in another major release

src/PublicApiGeneratorTests/Method_modifiers.cs
jnm2: Sounds good as-is 👍

danielmarbach: Seems microsoft mostly uses "static new"

https://github.com/dotnet/corefx/search?q=%22static+new%22&unscoped_q=%22static+new%22

I've asked on twitter. In the meantime let's pull this in as is. We can change it later once we know more

PR #155, opened 2019-11-21
Support for property modifiers

danielmarbach:
For now last bit of #112

### DIFF
--- a/src/PublicApiGenerator/CecilEx.cs
+++ b/src/PublicApiGenerator/CecilEx.cs
@@ -4,6 +4,7 @@ using System;
 using System.CodeDom;
 using System.Collections.Generic;
 using System.Linq;
+using Mono.Collections.Generic;
 
 namespace PublicApiGenerator
 {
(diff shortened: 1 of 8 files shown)


danielmarbach:
@jnm2 @sungam3r Last bit of #112 and I also did cleanup some redundancy given that the property name builder used the same switch expression (third time). This definitely increased the complexity of the code but I think it is still manageable and in the worst case the code can be inlined again if needed

src/PublicApiGenerator/CecilEx.cs
```diff
@@ -180,5 +181,24 @@ static CodeTypeReference ModifyCodeTypeReference(CodeTypeReference typeReference
...
+
+        internal static bool? IsNew<TDefinition>(this TDefinition methodDefinition, Func<TypeDefinition, Collection<TDefinition>?> selector, Func<TDefinition, bool> predicate)
+            where TDefinition : IMemberDefinition
+        {
+            bool? isNew = null;
+            var baseType = methodDefinition.DeclaringType.BaseType;
+            while (baseType is TypeDefinition typeDef)
+            {
+                isNew = selector(typeDef).Any(predicate);
+                if (isNew is true)
```
jnm2: `if (isNew)`?

jnm2: Never mind.

PR #163, opened 2020-01-13
Change default value from 'null' to 'default' for generic parameter

sungam3r:
Such a method

```c#
public TType Method<TType>(string name, TType defaultValue = default)
```

before this change was written as
```c#
public TType Method<TType>(string name, TType defaultValue = null)
```


sungam3r:
@danielmarbach Could you please review?

danielmarbach:
Sorry drunk in Mexico. Will review next week

sungam3r:
Would you publish 10.0.2 version on nuget?

danielmarbach:
It is on nuget now. Potentially still indexing but should be available soon

PR #166, opened 2020-01-28
Consistently write errors & log to standard error

atifaziz:
_Some_ errors would be sent to `STDOUT` as well as _all_ logging.

This PR consistently sends logging and errors to `STDERR` as opposed to having a mixed case of the two.

### DIFF
--- a/src/PublicApiGenerator.Tool/Program.cs
+++ b/src/PublicApiGenerator.Tool/Program.cs
@@ -11,8 +11,6 @@ namespace PublicApiGenerator.Tool
     /// </summary>
     public static class Program
     {
-        static TextWriter? _log;
-
         /// <summary>
         /// Public API generator tool that is useful for semantic versioning
         /// </summary>
@@ -43,12 +41,9 @@ namespace PublicApiGenerator.Tool
                 Path.Combine(workingDirectory, Path.GetRandomFileName()) :
                 Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
 
-            if (verbose)
-            {
-                _log = Console.Error;
-            }
+            var log = verbose ? Console.Error : null;
 
-            _log?.WriteLine($"Working area: {workingArea}");
+            log?.WriteLine($"Working area: {workingArea}");
 
             try
             {
@@ -56,11 +51,11 @@ namespace PublicApiGenerator.Tool
 
                 var template = CreateProjectTemplate(targetFrameworks, projectPath, package, packageVersion, generatorVersion!);
 
-                SaveProjectTemplate(workingArea, template, verbose);
+                SaveProjectTemplate(workingArea, template, log);
 
                 foreach (var framework in targetFrameworks.Split(";"))
                 {
-                    GeneratePublicApi(assembly, package, workingArea, framework, outputDirectory, verbose);
+                    GeneratePublicApi(assembly, package, workingArea, framework, outputDirectory, log);
                 }
 
                 return 0;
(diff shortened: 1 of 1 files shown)


atifaziz:
Before this PR, some errors and logging that would go to `STDOUT` would not get logged if `STDERR` was, for example, being redirected to a file.

danielmarbach:
I don't think it makes sense to write the verbose output to std error. I sent in an alternative that makes sure verbose output goes to Console and errors consistently go to std error

https://github.com/PublicApiGenerator/PublicApiGenerator/pull/167

atifaziz:
> I don't think it makes sense to write the verbose output to std error.

This is fairly conventional in command line applications. Verbose is considered diagnostics and that's what standard error is recommended for. You see, if you change the tool such that it writes the public API to standard output in some future version (that I'd like to propose :wink:) then logging to the same stream will corrupt the output.

danielmarbach:
> This is fairly conventional in command line applications. 

That is news to me. Would you mind to send in a few references so that I can educate myself?

By doing a quick look it seems most CLI tools from Microsoft use standard output (not even use STD ERROR)

https://github.com/dotnet/cli/issues/4910

> MSBuild does put everything on stdout. But I don't think that's the right design, and I'd rather go with the industry standard (primary output to stdout, warnings and errors to stderr) for any new tool.

> Every MS tool I can find for dotnet uses stdout for all output. Why have this tool be an outlier in our dotnet pipeline?

atifaziz:
> Would you mind to send in a few references so that I can educate myself?

I don't believe there is any authority on the topic or absolute rules but let's consider the wording in the POSIX standard as a fair source:

> At program start-up, three streams shall be predefined and need not be opened explicitly: _standard input_ (for reading conventional input), _standard output_ (for writing conventional output), and _standard error_ (for writing diagnostic output). When opened, the standard error stream is not fully buffered; the standard input and standard output streams are fully buffered if and only if the stream can be determined not to refer to an interactive device.
>
> ---
> Copyright © 2001-2018 IEEE and The Open Group, All Rights Reserved

Notice the bit reading “_standard error_ (for **writing diagnostic output**)” where it doesn't specifically limit standard error to error messages.

> By doing a quick look it seems most CLI tools from Microsoft use standard output

Perhaps most CLI tools you're looking into are build and SDK tools and so it's not surprising that most output naturally belongs on standard output.

I think the decision about what constitutes _standard output_ comes down to what the tool is designed to do and then everything else pretty much belongs on _standard error_. For example, [curl](https://curl.haxx.se/) only emits the HTTP response _body_ to standard output by default. That's what you would most naturally expect because you can harness the output by piping to other tools. If you ask for verbose output then it is emitted on standard error so that in-spite of it, the standard output of the program is the same. Another example is `git fetch` and `git fetch -v` where the latter emits the verbose output to _standard error_.

I see PublicApiGenerator as a generator so the generated public API text should be streamed to _standard output_. That way, one can pipe it to `grep` for searches, save to a file through redirection or what not. The verbose output seems to be about diagnostics like what all is happening behind the scenes (project, code, `dotnet run` output and so on) and so it seems to be a better fit for _standard error_. Above all, `--verbose` should not change the output. Also, should someone want it one stream then the user can, for example, redirect standard error to standard output (`2>&1`).

Right now, the PublicApiGenerator tool always generates the output to a file so I guess your _standard output_ is available for any purpose but I think that putting diagnostic information there means unnecessarily cornering yourself. This PR was about make it all consistent.

Issue #168, opened 2020-01-28
Wrong package version may be used silently (NU1603)


atifaziz:
I have been using this tool against packages coming off a build pipeline and I was getting puzzled on some occasions, and for quite some time, why the output would be from a different version than the expected one. After enabling verbose output, I finally discovered that warning [NU1603](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1603) was being issued along the lines of:

    'PackageA' 4.0.0 depends on 'PackageB' (>= 4.0.0) but 4.0.0 was not found. An approximate best match of 5.0.0 was resolved.

I propose a default behaviour of treating [warnings as errors](https://github.com/NuGet/Home/wiki/NuGet-Errors-and-Warnings) and a switch that suppresses warnings if someone wants to revert to the current behaviour.

danielmarbach:
Exact package version can be used, I updated the readme with a sample as part of #172

PR #169, opened 2020-01-28
Treat warnings as errors unless told otherwise via a CLI switch

atifaziz:
This PR is a proposal to address issue #168.

### DIFF
--- a/src/PublicApiGenerator.Tool/Program.cs
+++ b/src/PublicApiGenerator.Tool/Program.cs
@@ -19,10 +19,11 @@ namespace PublicApiGenerator.Tool
         /// <param name="generatorVersion">The version of the PublicApiGenerator package to use.</param>
         /// <param name="workingDirectory">The working directory to be used for temporary work artifacts. A temporary directory will be created inside the working directory and deleted once the process is done. If no working directory is specified the users temp directory is used.</param>
         /// <param name="outputDirectory">The output directory where the generated public APIs should be moved.</param>
+        /// <param name="suppressWarnings">Whether to suppress warnings (<c>true</c>) or treat them like errors (<c>false</c> and default).</param>
         /// <param name="verbose"></param>
         /// <param name="leaveArtifacts"></param>
         /// <returns></returns>
-        static int Main(string targetFrameworks, string? assembly = null, string? projectPath = null, string? package = null, string? packageVersion = null, string? generatorVersion = null, string? workingDirectory = null, string? outputDirectory = null, bool verbose = false, bool leaveArtifacts = false)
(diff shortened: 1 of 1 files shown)


danielmarbach:
If you switch the default behavior to be false (current behavior) I could even release it as a minor soonish. If we leave the default behavior as is I'd have to release it in a major which might delay this quite a bit.

atifaziz:
There are two ways to look at this: a bug or a new and incompatible behaviour. [Picking up the wrong version silently](https://github.com/PublicApiGenerator/PublicApiGenerator/issues/168) given the purpose of this tool seems incorrect (or surprising) and if we agree that it's a bug then I don't think anyone will be annoyed at it being rolled into a minor update. On the other hand, introducing a switch (like `--treat-warnings-as-errors`) to opt into what would yield least surprising behaviour seems odd and something you'd have to continue to opt into for the reminder of minor releases of version 1.

danielmarbach:
Well, warnings as errors is not even in csproj a default value. Given that the current behavior was designed to not treat warnings as errors and by default csproj behave the same and you have to opt-in for this I'm on the side of making this opt-in for the tool as well. If you don't want to do this that is totally fine. I will take your proposal, cherry-pick your commit and revert the default so that you don't have to bother with that

bording:
If the problem is specifically around using the wrong package version, then the project template could be modified to put brackets around the version number, which forces an exact version match, for example: https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/LibGit2Sharp.csproj#L30

If the exact package version can't be found, I believe that would then be an error, and not a warning.

atifaziz:
@bording That's an excellent suggestion and even simpler suggestion. It would just boil down to this:

```diff
diff --git a/src/PublicApiGenerator.Tool/Program.cs b/src/PublicApiGenerator.Tool/Program.cs
index f9d9cdd..d165568 100644
--- a/src/PublicApiGenerator.Tool/Program.cs
+++ b/src/PublicApiGenerator.Tool/Program.cs
@@ -228,7 +228,7 @@ private static string CreateProjectTemplate(string targetFrameworks, string? pro
 
         private static readonly string PackageReferenceTemplate =
             @"
-        <PackageReference Include=""{PackageName}"" Version=""{PackageVersion}"" />";
+        <PackageReference Include=""{PackageName}"" Version=""[{PackageVersion}]"" />";
 
         private static readonly string ProjectReferenceTemplate =
             @"<ItemGroup>
```

However, this is as sufficient a behavioural change as this PR so can't say if @danielmarbach would be comfortable with rolling it into a minor version update or not. Having said there, since the package version is never validated and stuck verbatim into a project file, there's nothing stopping one from taking today's version and specifying something like `--package-version [1.2.3]`. The following question would then be, is this going to be the supported way to specify versions going forward?

I would still say the an exact version match should be the default give that purpose of this tool. There should be no upgrades or downgrades happening automagically to avoid false positives for an approval.

atifaziz:
I am closing this PR on the basis that PR #170 supersedes it as a simpler solution (that doesn't require a new switch) thanks to @bording's [suggestion](https://github.com/PublicApiGenerator/PublicApiGenerator/pull/169#issuecomment-579445359).

PR #170, opened 2020-01-28
Use exact package version

atifaziz:
This PR implements [the suggestion](https://github.com/PublicApiGenerator/PublicApiGenerator/pull/169#issuecomment-579445359) from @bording in a [comment](https://github.com/PublicApiGenerator/PublicApiGenerator/pull/169#issuecomment-579445359) on PR #169. It either uses the exact version specified via the `--package-version` switch or fails. Like PR #169, this PR still constitutes a breaking behavioural change but I find it more of a bug fix (up to @danielmarbach to reconsider). Anyone relying on the current behaviour could be mislead (the chance that the public API generated is from another version than the one specified), as I was and which prompted the reporting of issue #168.

### DIFF
--- a/src/PublicApiGenerator.Tool/Program.cs
+++ b/src/PublicApiGenerator.Tool/Program.cs
@@ -228,7 +228,7 @@ namespace PublicApiGenerator.Tool
 
         private static readonly string PackageReferenceTemplate =
             @"
-        <PackageReference Include=""{PackageName}"" Version=""{PackageVersion}"" />";
+        <PackageReference Include=""{PackageName}"" Version=""[{PackageVersion}]"" />";
 
         private static readonly string ProjectReferenceTemplate =
             @"<ItemGroup>


danielmarbach:
I had a closer look at this and I'm not going to pull this in as of now. Instead I updated the readme to show how exact version match can be used

https://github.com/PublicApiGenerator/PublicApiGenerator/pull/172

Already before that PR the readme pointed for example that wildcards are supported

> ```
> --package-version Version
> ```
> 
> A nuget package version or floating versions as specified by https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files. For example
> 
> - 8.1.1
> - 8.0
> - 8.0-*

Making it by default use `[]` (exact match) would break that. Thanks again for the PR and for the effort you put into this!

atifaziz:
@danielmarbach That's a fair call.

> Already before that PR the readme pointed for example that wildcards are supported
> 
> > ```
> > --package-version Version
> > ```
> > 
> > 
> > A nuget package version or floating versions as specified by https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files. For example
> > 
> > * 8.1.1
> > * 8.0
> > * 8.0-*

While that's true, the referenced page doesn't mention anything about ranges whatsoever; it's [found under package versioning section of the NuGet docs](https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges-and-wildcards). Now that it's in this tool's README (thanks), it's quite clear that it's possible and supported.

danielmarbach:
Yep very great input. Another case of "it is only clear to the person who wrote it" :D

Thanks for all your helpful suggestions

PR #177, opened 2020-01-29
Spike (2) generating API to STDOUT

atifaziz:
@danielmarbach I am sharing a second tool that I started in the same spirit of addressing #173 before it was opened as an issue. I am sharing it as a spike in this PR. The reason I added a second tool is that I wasn't very hopeful that all the ideas would be welcome (sensing, perhaps falsely, some resistance with the other PRs I submitted) so I wanted a branch in my fork that I could always rebase or merge to keep up with the changes in this project. After all, the tools do very little and the bulk of the API generation is in the library so it seemed like an acceptable compromise in the interest of time. If you like what you see, I'd be happy to contribute.

I have done some other minor refactorings and changes to improve the output but those are completely secondary.

You are free to close this PR and I am happy to maintain my own fork but I'd rather collaborate and contribute back if the feedback and ideas are welcome.

### Multiple target frameworks

```
dotnet run -p .\src\PublicApiGenerator.Cli -- --package morelinq --package-version [3.3.0] --target-frameworks 'netcoreapp3.1;netcoreapp2.1'
```

output:

```
A:\PublicApiGenerator\morelinq.netcoreapp3.1.received.txt
A:\PublicApiGenerator\morelinq.netcoreapp2.1.received.txt
```

### Single target framework + `grep`

```
dotnet run -p .\src\PublicApiGenerator.Cli -- --package morelinq --package-version [3.3.0] --target-frameworks netcoreapp2.1 | grep Zip
```

output:

```
    public static class BatchExtension
        public static System.Collections.Generic.IEnumerable<System.Collections.Generic.IEnumerable<TSource>> Batch<TSource>(this System.Collections.Generic.IEnumerable<TSource> source, int size) { }
        public static System.Collections.Generic.IEnumerable<TResult> Batch<TSource, TResult>(this System.Collections.Generic.IEnumerable<TSource> source, int size, System.Func<System.Collections.Generic.IEnumerable<TSource>, TResult> resultSelector) { }
        public static System.Collections.Generic.IEnumerable<System.Collections.Generic.IEnumerable<TSource>> Batch<TSource>(this System.Collections.Generic.IEnumerable<TSource> source, int size) { }
        public static System.Collections.Generic.IEnumerable<TResult> Batch<TSource, TResult>(this System.Collections.Generic.IEnumerable<TSource> source, int size, System.Func<System.Collections.Generic.IEnumerable<TSource>, TResult> resultSelector) { }
```

PR #178, opened 2020-01-30
Refactor project creation using strong XML model

atifaziz:
With this PR (borrowed from #177), I'd like to propose changing the project XML string replacements (that's hard to read and maintain) with strong-typed XML tree building. It also has the benefit of replacing 3 intertwined string fields (`ProjectTemplate`, `PackageReferenceTemplate` and `ProjectReferenceTemplate`) with a single and self-contained function that is now `SaveProject`.

I have also renamed `CreateProjectTemplate` and `SaveProjectTemplate` to `CreateProject` and `SaveProject` since they deal with the _actual_ project rather than a project's template.

### DIFF
--- a/src/PublicApiGenerator.Tool/Program.cs
+++ b/src/PublicApiGenerator.Tool/Program.cs
@@ -4,6 +4,7 @@ namespace PublicApiGenerator.Tool
     using System.Diagnostics;
     using System.IO;
     using System.Reflection;
+    using System.Xml.Linq;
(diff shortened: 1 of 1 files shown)


danielmarbach:
Haha interesting from me this variant is harder to read but I can see your arguments and happy to pull it in.

atifaziz:
Appreciate you merging in spite of finding the XML tree-building harder to read.

PR #179, opened 2020-01-30
Embed sub-program source as a resource

atifaziz:
Borrowing from #177, I'd like to propose moving the generator sub-program source code into a separate file that's embedded as a resource as opposed to a string literal to help with maintenance, validation and easier reading (syntax highlighting).

Also note that I removed the unnecessary `public` modifiers from the sub-program's `Program` and `Main` (thus defaulting to `internal`). If you prefer to record that as a separate edit/commit then I can split it from this PR.


danielmarbach:
Needs a rebase

danielmarbach:
Great addition and I agree the previous approach wasn't ideal from the perspective you highlighted

Issue #173, opened 2020-01-29
Write API output to STD out


danielmarbach:
Idea from @atifaziz 

> You see, if you change the tool such that it writes the public API to standard output in some future version (that I'd like to propose 😉)

danielmarbach:
@atifaziz do you have have a recommendation on how the tool should handle API outputs to multiple target frameworks?

Currently, the tool creates a file per target framework. Would a `Console.WriteLine` per target framework be enough?

atifaziz:
> do you have have a recommendation on how the tool should handle API outputs to multiple target frameworks?

The tool would need two modes of operation:

1. Generate API to standard output if only a single target framework is given and the `--output-directory` argument is absent.
2. Generate files per framework API like now but print the generate file paths to standard output.

If the `--output-directory` argument is supplied then it's clear that the user wants a file generated even if a single target framework is given, and in which case, the tool should enter mode 2.

However, I would argue that we shouldn't even bother with multiple target frameworks. If the tool just works with one target framework then the user can just use loop from the shell to invoke the tool for each framework and redirect the output to file if they prefer file generation:

Windows example:

```cmd
for %f in (netcoreapp2.1 netcoreapp3.1) do ^
  generate-public-api --package foobar --package-version 42.0.0 --target-framework %f ^
    > foobar.%f.received.txt
```

PowerShell example:

```ps1
'netcoreapp2.1', 'netcoreapp3.1' |
  % { generate-public-api --package foobar --package-version 42.0.0 --target-framework $_ |
        Out-File "foobar.$f.received.txt" -Encoding utf8 }
```

Bash example:

```sh
for f in netcore2.1 netcore3.1; do
  generate-public-api --package foobar --package-version 42.0.0 --target-framework ${f} \
    > foobar.${f}.received.txt
done
```

In essence, the tool would become very simple and just do one thing and do it well.

The only argument I see in favour of supporting a multiple-target-frameworks argument is that you want the tool invocation to be to be independent of shells (use doesn't have to adapt to the looping syntax in each shell) so you make looping through target frameworks a native feature (but as a nice-to-have and a secondary mode of operation). It does however complicate things by introducing multiple modes of operation.

danielmarbach:
It sort of has to be independent of shell because it is shipped as a dotnet global tool to essentially any environment that supports .NET Core and global tools.

atifaziz:
[...]

PR #180, opened 2020-01-30
Generate API on STDOUT when single target framework given

atifaziz:
This PR addresses #173.

## Notable Changes

The _worker_ project's `Program.Main` takes two parameters only: the source assembly path and the output API file path. If the latter is `-` then the API is generated to standard output. It also logs unhandled exceptions to standard error and sets an exit code on success or failure. It is no longer the responsibility of the worker project to do any file operations like deleting and moving. This is done by PublicApiGenerator.

If `dotnet run` times out (after 10 seconds, which seems short to be honest) then an exception is thrown. Previously, PublicApiGenerator continue silently.

### DIFF
--- a/src/PublicApiGenerator.Tool/Program.cs
+++ b/src/PublicApiGenerator.Tool/Program.cs
@@ -3,6 +3,7 @@ namespace PublicApiGenerator.Tool
     using System;
     using System.Diagnostics;
     using System.IO;
+    using System.Linq;
     using System.Reflection;
     using System.Text;
     using System.Xml.Linq;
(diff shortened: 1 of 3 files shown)


src/PublicApiGenerator.Tool/Program.cs
```diff
@@ -86,50 +89,104 @@ private static void GeneratePublicApi(string? assembly, string? package, string
...
+            foreach (var (i, arg) in arguments.Select((arg, i) => (i + 1, arg)))
+            {
+                logVerbose.WriteLine($"{i,4}. {arg}");
+                psi.ArgumentList.Add(arg);
+            }
             logVerbose.WriteLine();
 
+            using var process = Process.Start(psi);
+
+            static DataReceivedEventHandler
```
sungam3r: The local function mixes with the body of the method and looks wild.

atifaziz: Hope it looks better after fdf6d1a471429be2b0d9934338c7cdb998929872.

danielmarbach:
Replaced by https://github.com/PublicApiGenerator/PublicApiGenerator/pull/181 which contains just a minor cleanup

PR #181, opened 2020-01-31
Generate API on STDOUT when single target framework given

danielmarbach:
This PR addresses #173.
Replaces #180 

## Notable Changes

The _worker_ project's `Program.Main` takes two parameters only: the source assembly path and the output API file path. If the latter is `-` then the API is generated to standard output. It also logs unhandled exceptions to standard error and sets an exit code on success or failure. It is no longer the responsibility of the worker project to do any file operations like deleting and moving. This is done by PublicApiGenerator.

If `dotnet run` times out (after 10 seconds, which seems short to be honest) then an exception is thrown. Previously, PublicApiGenerator continue silently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,18 @@ Generate public API for fluent assertions 5.* for runtime framework `net47`
 generate-public-api --target-frameworks net47 --package FluentAssertions --package-version 5.*
 ```
 
+Note that when a single target framework is specified then the API is generated to standard output. To write to a file, you can either use shell redirection:
+
+```
+generate-public-api --target-frameworks net47 --package FluentAssertions --package-version 5.* > api.txt
+```
+
+or specify an output directory to force the generation of an API file:
+
+```
+generate-public-api --target-frameworks net47 --package FluentAssertions --package-version 5.* --output-directory .
+```
+
 Generate public API for fluent assertions 5.6.0 (exact version match) for runtime framework `net47`
 
 ```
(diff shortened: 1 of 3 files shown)


sungam3r:
> If dotnet run times out (after 10 seconds, which seems short to be honest) then an exception is thrown. 

Just one note. I have one project (assembly) for which public API is generated in more than 50 seconds.

danielmarbach:
I think I'll set the default timeout to 60 seconds and make it configurable as part of the minor. Will do after the weekend -------- Ursprüngliche Nachricht --------Von: Ivan Maximov <notifications@github.com>Datum: Sa., 1. Feb. 2020, 07:47An: PublicApiGenerator/PublicApiGenerator <PublicApiGenerator@noreply.github.com>Cc: Daniel Marbach <daniel.marbach@openplace.net>, State change <state_change@noreply.github.com>Betreff: Re: [PublicApiGenerator/PublicApiGenerator] Generate API on STDOUT when single target framework given (#181)
If dotnet run times out (after 10 seconds, which seems short to be honest) then an exception is thrown.

Just one note. I have one project (assembly) for which public API is generated in more than 50 seconds.

danielmarbach:
@sungam3r what do you think about https://github.com/PublicApiGenerator/PublicApiGenerator/pull/185 ?

sungam3r:
LGTM

Issue #188, opened 2020-05-16
Indexer arguments are lost if indexer is abstract


stakx:
This appears to be a regression introduced in 0cca8a8a5ab18e089a01ca756364a891b0b6eddc. I noticed it when upgrading from version 9.3.0 (working) to 10.0.2 (not working anymore).

#### Repro code:

```csharp
using System;
using System.Reflection;

using PublicApiGenerator;

public abstract class C1
{
    public abstract object this[object key] { get; set; }
}

public class C2
{
    public object this[object key] { get => throw null; set => throw null; }
}

public interface I
{
    object this[object key] { get; set; }
}

class Program
{
    static void Main()
    {
        Console.WriteLine(Assembly.GetExecutingAssembly().GeneratePublicApi());
    }
}
```

#### Actual result:

Outputs the following public API definition:

```csharp
[assembly: System.Runtime.InteropServices.ComVisible(false)]
[assembly: System.Runtime.InteropServices.Guid("95e17962-cb9d-4b92-a2cc-20c8f70c7f80")]
[assembly: System.Runtime.Versioning.TargetFramework(".NETFramework,Version=v4.6.2", FrameworkDisplayName=".NET Framework 4.6.2")]
public abstract class C1
{
    protected C1() { }
    public abstract object Item { get; set; }
}
public class C2
{
    public C2() { }
    public object this[object key] { get; set; }
}
public interface I
{
    object this[object key] { get; set; }
}
```

The output for `C1.Item` is incorrect: the indexer arguments aren't there.

#### Expected result:

The indexer in C1 should have been output as such:

```diff
-    public abstract object Item { get; set; }
+    public abstract object this[object key] { get; set; }
```

#### Preliminary analysis:

I suppose that the name change from `Item` to `_ abstract _5DB9F56043FF464997155541DA321AD4_Item` caused by this:

https://github.com/PublicApiGenerator/PublicApiGenerator/blob/4dfb9f65b6b6f5159daf93cd318e34a17ed4907e/src/PublicApiGenerator/ModifierMarkerNameBuilder.cs#L19

then causes CodeDOM to emit incorrect code? If so, I suppose that's what the source code is referring to here:

https://github.com/PublicApiGenerator/PublicApiGenerator/blob/4dfb9f65b6b6f5159daf93cd318e34a17ed4907e/src/PublicApiGenerator/ApiGenerator.cs#L737

PR #189, opened 2020-05-16
Fix output for abstract get-set indexers

stakx:
Tests for, and fixes, #188... but I don't know what collateral damage this may produce, as my understanding of `PropertyNameBuilder.AugmentPropertyNameWithPropertyModifierMarkerTemplate` is rather incomplete. While all existing tests still pass, I could easily be breaking some yet untested use cases.


danielmarbach:
Hi, thanks for the contribution! I've added as many tests I could think off and yet still we sometimes discover cases we haven't thought off. Don't worry if something falls between the cracks we can always release another fix

danielmarbach:
Released as part of 10.1.0, thanks again for the help!

stakx:
Awesome! Thank you @danielmarbach.

PR #191, opened 2020-06-17
Ensure that `SerializableAttribute` gets reproduced in output

### LINKED ISSUE #190: [Serializable] attribute does not show up in contracts
```csharp
[Serializable] class SomeException : Exception { }
```

PublicApiGenerator won't reproduce the `[Serializable]` attribute in the generated API string&mdash;assuming default `PublicApiGeneratorOptions` which don't exclude it.

This is likely due to `SerializableAttribute` being a "pseudo" custom attribute that gets transformed to a `TypeAttributes` flag in the IL metadata. Unlike System.Reflection, Mono Cecil is more faithful to the actual metadata and won't report it as a custom attribute when querying `type.CustomAttributes` [here](https://github.com/PublicApiGenerator/PublicApiGenerator/blob/a2a24907603fc853cae38f03e18923fa0d461f85/src/PublicApiGenerator/ApiGenerator.cs#L439).

stakx:
Repro unit test and fix for #190.

~~Marking this as a draft, as I might add more commits later and convert this into a bugfix PR.~~

Fixes #190.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -441,6 +441,23 @@ namespace PublicApiGenerator
                 var attribute = GenerateCodeAttributeDeclaration(codeTypeModifier, customAttribute);
                 attributes.Add(attribute);
             }
+
+            // so this is not very cool but at least all the attribute creation is in the same place
+            PopulateAttributesThatDontAppearInCustomAttributes(type, attributes);
+        }
+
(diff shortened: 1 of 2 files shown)


danielmarbach:
@stakx have a look. I only added handling for the `SerializableAttribute` for now. It is not the most beautiful solution but the attribute population is quite white spread so I decided to put the special case handling for type declarations close to the attribute population even though it requires safe casting. With that at least the attribute creation logic is somehow close to each other.

PR #193, opened 2020-06-17
Fix syntax for overloaded operator methods

### LINKED ISSUE #192: Syntactically incorrect output for overloaded operator methods
I just stumbled on two small issues with overloaded operators:

* The `operator` keyword is missing in the generated output.
* `op_False` / `op_True` don't get mapped to the `false` / `true` keywords.

I'll submit a PR for this shortly.

stakx:
Follow-up to #49. Fixes #192.


danielmarbach:
Thanks for finding and fixing it!

danielmarbach:
Should go out in a few mins as 10.1.2

sungam3r:
I never cease to be surprised at how we could have missed so many (obvious) details.

PR #217, opened 2020-12-06
Ignore JetBrains Rider generated files

MaxShoshin:


danielmarbach (approved):

stakx (approved):
Fine by me, but I'm not using Rider regularly so I cannot judge whether this is correct. I'm happy to trust you, though. :-)

(Solely for reference, [this JetBrains support page](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems) states this:

> Here is what you need to share:
>
> * All the files under the `.idea` directory in the project root except the `workspace.xml`, `usage.statistics.xml`, and `tasks.xml` files and the `shelf` directory which store user-specific settings
> * All the `.iml` module files that can be located in different module directories (applies to IntelliJ IDEA)

But I can't say whether that is also valid for Rider.)

Issue #213, opened 2020-11-16
Generate API for internal members


MaxShoshin:
Currently there is no way to include in generated result internal classes.

I understand, that the internal members are not Public API. 
But may be create special setting to do not filter out internal members? 

It seems that sometimes it will be usefull.

My use case: 
I try to use PublicApiGenerator to support code review process on changes in my serialization/deserialization process. 
I would like to iterate all members used in serialization/deserialization and if something changes - I put additional attention to have these changes in serialization api to be backward compatible. 

In my case I serialize/deserialize internal types. And I cannot use PublicApiGenerator to generate API used during serialization - it filtered out internal classes.

And I don't want to expose them as public - to avoid access to my classes from other assemblies.

What do you think about adding special settings to include internal members? Or may be you have some kind of woraround for my case?

Thank you!

stakx:
I would principally argue that you are trying to use this project for something for which it clearly wasn't designed, and it probably doesn't make sense to extend it in your desired direction.

That being said, I'd also admit that this request _might_ be sensible in certain situations where `[assembly: InternalsVisibleTo(...)]` attributes are involved. Say you want to generate the public API for assembly A. You're invoking PublicApiGenerator from assembly G. Assembly A carries a `[assembly: InternalsVisibleTo("G")]` attribute, meaning that A's API surface visible to G encompasses more than just `public` members. In such a situation, a flag on `ApiGeneratorOptions` determining whether or not to include non-public members might make sense.

danielmarbach:
The thing is though if you have InternalsVisibleTo you are no longer the public API you are coupling yourself to the internal API. Given that this tool is called PublicApiGenerator I would argue that it was never the intention to verify internal APIs (not even if you have internals visible to)

MaxShoshin:
Ok.
I agree that internal api is not public api. 

But...

May be we can split: api generation (regardless visibility) and facade which working only with public api?

I really would like use this awesome project to generate text from api surface.

In my particular case I use 'internal' modifier only as technical trick to avoid made public a lot of other related classes. But in the meaning, classes which are used to serialize/deserialize data (json in my case) - are public api: if you change them - you can break functionality, break compatibility.

I'm working directly with `GenerateApi` class to generate public api text, so it will be nice if I can "override" `ShouldIncludeType(t)` method in following condition:
[...]

PR #218, opened 2020-12-06
Pull up type condition (which type will be included in result)

MaxShoshin:
This change help me to workaround for #213 (using reflection, using my custom filtration, on my own risk, but I can be able to do it).
Also it doesn't expose ability to generate api text for internal member.

Also this change will get code bit more cleaner (single place for filter types) (I hope).

This change join two predicates which are used to determine which types will be included in result into single one.
Originally part of the predicate was in `GeneratePublicApi` method, another part - in `CreatePublicApiForAssembly` (internal method). 
Now method `CreatePublicApiForAssembly` will receive full predicate from method `GeneratePublicApi`.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -45,7 +45,9 @@ namespace PublicApiGenerator
                 {
                     return CreatePublicApiForAssembly(
                         asm,
-                        typeDefinition => options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName),
+                        typeDefinition =>!typeDefinition.IsNested &&
+                                         ShouldIncludeType(typeDefinition) &&
+                                         (options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName)),
                         options.IncludeAssemblyAttributes,
                         options.WhitelistedNamespacePrefixes,
                         attributeFilter);
@@ -111,7 +113,7 @@ namespace PublicApiGenerator
                 }
 
                 var publicTypes = assembly.Modules.SelectMany(m => m.GetTypes())
-                    .Where(t => !t.IsNested && ShouldIncludeType(t) && shouldIncludeType(t))
+                    .Where(shouldIncludeType)
                     .OrderBy(t => t.FullName, StringComparer.Ordinal);
                 foreach (var publicType in publicTypes)
                 {
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -45,9 +45,9 @@ namespace PublicApiGenerator
                 {
                     return CreatePublicApiForAssembly(
                         asm,
-                        typeDefinition =>!typeDefinition.IsNested &&
-                                         ShouldIncludeType(typeDefinition) &&
-                                         (options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName)),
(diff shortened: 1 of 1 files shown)


danielmarbach:
I'll leave it up to @stakx to decide whether he wants to take this or not. I already shared my opinion on the other thread. 

Putting my doubts aside for a second, would it make sense then to have the default delegate declared on the option and then allow providing a different impl over the option?

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -45,7 +45,9 @@ public static string GeneratePublicApi(this Assembly assembly, ApiGeneratorOptio
                 {
                     return CreatePublicApiForAssembly(
                         asm,
-                        typeDefinition => options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName),
+                        typeDefinition =>!typeDefinition.IsNested &&
+                                         ShouldIncludeType(typeDefinition) &&
+                                         (options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName)),
```
sungam3r, line 50:
```suggestion
                        typeDefinition => !typeDefinition.IsNested &&
                                          ShouldIncludeType(typeDefinition) &&
                                          (options.IncludeTypes == null || options.IncludeTypes.Any(type => type.FullName == typeDefinition.FullName && type.Assembly.FullName == typeDefinition.Module.Assembly.FullName)),
```

MaxShoshin:
Hm... 

As user - I prefer to be able to overrider behavior via options (to be able to solve my issue in the easiest way). 
But predicate "as is" contains reference to Mono.Cecil types. I have some doubts to have reference to the third party library in public API of this library (may be use `Func<Type, bool>`?).

stakx:
This looks so deceptively unrelated to the original issue that I didn't quite trust its simplicity initially, and I had to stare at it for several minutes, thinking about how this could possibly help you. But I'm taking your word for it that it does. At the end of the day, it's a simple change &ndash; moving two conditions from one method to another (which gets called exclusively by the former), so it should be harmless.

I'm willing to merge this PR mostly because I tend to agree that it's indeed _slightly_ neater to concentrate that query in a single method instead of two. (Of course we could possibly do similar simplifications elsewhere, but that's not the goal here.)

That being said...

> This change help me to workaround for #213 (using reflection, using my custom filtration, on my own risk, but I can be able to do it).

Frankly, I think you would do better to fork this project and make any necessary changes towards your goal directly in source code, instead of relying on reflection tricks and workarounds. Especially if your current approach involves internal implementation details of PublicApiGenerator, it could stop working at any future moment. It might be safer, and also more transparent, to sync your fork with its upstream (i.e. this repository) whenever required.

If there are going to be further PRs towards enabling your particular use case, I think we should stop right here. However, if this is the only change required, we're good to go.

stakx:
> Putting my doubts aside for a second, would it make sense then to have the default delegate declared on the option and then allow providing a different impl over the option?

@danielmarbach, I share your doubts. Those can certainly be put aside purely for discussion's sake, but eventually we'll still have to decide whether or not we should allow inclusion of non-public types in the generated API. I could well imagine that this will lead to new bugs surfacing &ndash; code has been written with the assumption that it only has to deal with public types. Would we want to deal with such bugs (_if_ they actually surface)?

My vote goes towards keeping PublicApiGenerator true to its name... for now. If more people show up supporting the scenario described in #213, making the type selection predicate overridable via `options` would seem like a good idea. 

> As user - I prefer to be able to overrider behavior via options (to be able to solve my issue in the easiest way).
But predicate "as is" contains reference to Mono.Cecil types. I have some doubts to have reference to the third party library in public API of this library (may be use Func<Type, bool>?).

Good point. I agree that ideally we ought to treat Mono.Cecil like an implementation detail and not expose it through our own API.

stakx (approved):

MaxShoshin:
> If there are going to be further PRs towards enabling your particular use case, I think we should stop right here.
I don't expect any additional PRs regarding my particular use case. I agree with you about doubts regarding name.

PR #226, opened 2021-06-08
[DEMO] Add test for new modifier

sungam3r:
`MetadataResolver.GetMethod(d.Methods, method)` from `static bool IsHidingMethod(MethodDefinition method)` returns `null` . I do not know whether this a bug or System.CodeDom limitation.


stakx:
> `MetadataResolver.GetMethod(d.Methods, method)` from `static bool IsHidingMethod(MethodDefinition method)` returns `null` . I do not know whether this a bug or System.CodeDom limitation.

This appears to be a problem in PublicApiGenerator's own `CecilEx.IsHidingMethod`: `MetadataResolver.GetMethod` takes return types into account ([here](https://github.com/jbevain/cecil/blob/634a37d3f2be7b8c72a612fce7b4907a12533d4c/Mono.Cecil/MetadataResolver.cs#L264-L265)); however, those aren't actually relevant when it comes to method shadowing / hiding. Therefore the use of `MetadataResolver.GetMethod` in `IsHidingMethod` is probably inappropriate. We should perhaps look for a different API to use.

sungam3r:
Thank you, it is clear now.

stakx:
@sungam3r, apologies once more for ignoring this PR for so long. Would you still like to continue work on this? If so, may I ask that you rebase this on top of current `master`?

If I remember correctly, the test that you're adding actually fails, so obviously we shouldn't merge the PR without also adding a fix for the failing test.

sungam3r:
Done. Yes, this is just a test to demonstrate problem.

stakx:
> Yes, this is just a test to demonstrate problem.

Oh, that's fine. Could you please convert this PR to a draft, so we don't accidentally merge it? 🙃

PR #247, opened 2022-08-11
Fix inconsistencies for record types across NET 6.x.x SDK

### LINKED ISSUE #245: API string generated differently on Linux vs Windows under 6.0.400
## Issue

Under 6.0.400, record types cause API strings to generate with this on Linux:

```cs
protected virtual System.Type EqualityContract { get; }
```

However, on Windows the above is not listed within the record type definition.

In our case, the record is defined as this:

```cs
public record ExecutionStrategyRegistration(IExecutionStrategy Strategy, OperationType Operation);
```

where `OperationType` is an enum.

Generating the API file under 6.0.303 does not produce the `EqualityContract` within the API file.

## Expected

The generated API should be the same cross-platform -- either with `EqualityContract` or without.

sungam3r:
fixes #245

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-        sdk: ["5.0.408", "6.0.302", "6.0.303", "6.0.400"]
+        sdk: ["6.0.100", "6.0.302", "6.0.303", "6.0.400"]
     name: ${{ matrix.runs-on }}-${{ matrix.sdk }}
     runs-on: ${{ matrix.runs-on }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,10 +12,14 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-        sdk: ["3.1.422", "5.0.408", "6.0.302", "6.0.303", "6.0.400"]
+        sdk: ["5.0.408", "6.0.302", "6.0.303", "6.0.400"]
     name: ${{ matrix.runs-on }}-${{ matrix.sdk }}
     runs-on: ${{ matrix.runs-on }}
     steps:
+    - name: Setup .NET Core 3.1 SDK # should always be installed for 3.1 runtime
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: 3.1.422
     - name: Setup .NET Core SDK
       uses: actions/setup-dotnet@v2
       with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,9 @@ jobs:
     - name: Setup .NET Core SDK
       uses: actions/setup-dotnet@v2
       with:
-        dotnet-version:
-        - 3.1.422
-        - ${{ matrix.sdk }}
+        dotnet-version: |
+          3.1.422
+          ${{ matrix.sdk }}
     - run: dotnet --info
     - uses: actions/checkout@af513c7a016048ae468971c52ed77d9562c7c819
     - if: contains(matrix.runs-on, 'windows')
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-        sdk: ["6.0.100", "6.0.302", "6.0.303", "6.0.400"]
+        sdk: ["6.0.202", "6.0.302", "6.0.303", "6.0.400"]
     name: ${{ matrix.runs-on }}-${{ matrix.sdk }}
     runs-on: ${{ matrix.runs-on }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-        sdk: ["6.0.202", "6.0.302", "6.0.303", "6.0.400"]
+        sdk: ["6.0.301", "6.0.302", "6.0.303", "6.0.400"]
     name: ${{ matrix.runs-on }}-${{ matrix.sdk }}
     runs-on: ${{ matrix.runs-on }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,4 +24,4 @@ jobs:
     - uses: actions/upload-artifact@master
       with:
         name: nugets
-        path: nugets/  
+        path: nugets/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,9 +12,14 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-    name: ${{ matrix.runs-on }}
+        sdk: ["6.0.302", "6.0.303", "6.0.400"]
+    name: ${{ matrix.runs-on }}-${{ matrix.sdk }}
     runs-on: ${{ matrix.runs-on }}
     steps:
+    - name: Setup .NET Core SDK
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: ${{ matrix.sdk }}
     - run: dotnet --info
     - uses: actions/checkout@af513c7a016048ae468971c52ed77d9562c7c819
     - if: contains(matrix.runs-on, 'windows')
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-        sdk: ["6.0.302", "6.0.303", "6.0.400"]
+        sdk: ["3.1.422", "5.0.408", "6.0.302", "6.0.303", "6.0.400"]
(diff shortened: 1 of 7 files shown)


sungam3r:
@Shane32 3.1 runtime fails all tests, 6.0.400 success on both Windows and Linux!

sungam3r:
It seems that there are some inconsistencies between CompilerGenerated attributes accross SDKs. Nevermind, I just ignored `EqualityContract` member to unify behavior. Also NET5 (and even NET 6.0.100) has completely different output from NET6:
```
 public class User : System.IEquatable<PublicApiGeneratorTests.Examples.User>
    {
        protected User(PublicApiGeneratorTests.Examples.User original) { }
        public User(string login, string password) { }
        public string login { get; set; }
        public string password { get; set; }
        public virtual PublicApiGeneratorTests.Examples.User <Clone>$() { }
        public void Deconstruct(out string login, out string password) { }
        public virtual bool Equals(PublicApiGeneratorTests.Examples.User? other) { }
        public override bool Equals(object? obj) { }
        public override int GetHashCode() { }
        protected virtual bool PrintMembers(System.Text.StringBuilder builder) { }
        public override string ToString() { }
        public static bool operator !=(PublicApiGeneratorTests.Examples.User? left, PublicApiGeneratorTests.Examples.User? right) { }
        public static bool operator ==(PublicApiGeneratorTests.Examples.User? left, PublicApiGeneratorTests.Examples.User? right) { }
```
vs
```
public class User : System.IEquatable<PublicApiGeneratorTests.Examples.User>
    {
        public User(string login, string password) { }
        public string login { get; set; }
        public string password { get; set; }
    }
```

Shane32:
I think I remember a .NET compiler bug saying that the generated members were not marked as compiler generated or something.  I don't have a reference, but chances are the bug was resolved after 6.0.100 and hence the difference.

sungam3r:
@stakx I will be grateful to you for review/publish the fix.

stakx:
@sungam3r, I'm about to go on vacation for a few weeks' time. If time permits, I'll try to review this before I'm off.

Shane32:
@stakx It would be greatly appreciated

.github/workflows/ci.yml
```diff
@@ -12,9 +12,18 @@ jobs:
...
     runs-on: ${{ matrix.runs-on }}
     steps:
+    - name: Setup .NET Core 3.1 SDK # should always be installed for 3.1 runtime
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: 3.1.422
+    - name: Setup .NET Core SDK
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: ${{ matrix.sdk }}
```
Shane32, line 26:
```suggestion
    - name: Setup .NET Core SDK
      uses: actions/setup-dotnet@v2
      with:
        dotnet-version:
        - 3.1.422
        - ${{ matrix.sdk }}
```
@sungam3r does this work?

sungam3r: Probably, let's try

sungam3r:
Also affects https://github.com/EasyNetQ/EasyNetQ/pull/1441

.github/workflows/ci.yml
```diff
@@ -12,9 +12,16 @@ jobs:
...
+        sdk: ["6.0.301", "6.0.302", "6.0.303", "6.0.400"]
+    name: ${{ matrix.runs-on }}-${{ matrix.sdk }}
     runs-on: ${{ matrix.runs-on }}
     steps:
+    - name: Setup .NET Core SDK
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version:
+        - 3.1.422
+        - ${{ matrix.sdk }}
```
Shane32, line 24: I did it wrong.  I think it's supposed to be like this:

```suggestion
        dotnet-version: |
          3.1.422
          ${{ matrix.sdk }}
```

.github/workflows/ci.yml
```diff
@@ -12,9 +12,16 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest, windows-latest]
-    name: ${{ matrix.runs-on }}
+        sdk: ["6.0.301", "6.0.302", "6.0.303", "6.0.400"]
```
stakx, line 15: Out of curiosity, what is the significance of running CI builds for several patch versions of .NET 6? Why not just pick the most recent one of those listed here?

src/PublicApiGeneratorTests/ApiGeneratorTestsBase.cs
```diff
@@ -34,8 +34,19 @@ private static void AssertPublicApi(Assembly assembly, string expectedOutput,  A
...
+                    ignoreWhiteSpaceDifferences: true);
+            }
+            catch
+            {
+                Console.WriteLine("Full expected output:");
+                Console.WriteLine(expectedOutput);
+                Console.WriteLine("Full actual output:");
+                Console.WriteLine(actualOutput);
+                throw;
+            }
```
stakx, line 49: I don't fully understand why this is needed. Shouldn't `Assert.Equal` already log expected vs. actual values when they differ? Using `Console.WriteLine` durring unit test execution doesn't seem like the cleanest thing to do...

src/PublicApiGeneratorTests/Assembly_attributes.cs
```diff
@@ -21,6 +21,23 @@ public class Assembly_attributes : ApiGeneratorTestsBase
...
         {
+#if NET6_0
+            var api = @"[assembly: PublicApiGeneratorTests.Examples.AttributeWithMultiplePositionalParameters(42, ""Hello"")]
+[assembly: PublicApiGeneratorTests.Examples.AttributeWithNamedParameter(IntValue=42, StringValue=""Hello"")]
+[assembly: PublicApiGeneratorTests.Examples.AttributeWithPositionalParameters1(""Hello"")]
+[assembly: PublicApiGeneratorTests.Examples.AttributeWithPositionalParameters2(42)]
+[assembly: PublicApiGeneratorTests.Examples.Simple]
+[assembly: System.Runtime.InteropServices.ComVisible(false)]
+[assembly: System.Runtime.InteropServices.Guid(""3B8D506A-5247-47FF-B053-D29A51A97C33"")]
+[assembly: System.Runtime.Versioning.TargetFramework("".NETCoreApp,Version=v6.0"", FrameworkDisplayName="""")]
```
stakx, line 32: There's a lot of duplication being introduced here; took me a while to see that the only difference appears to be with the first argument to the `TargetFramework` attribute. Perhaps that value could have been extracted along the following line:

```csharp
#if NET6_0
    const string TFM = ".NETCoreApp,Version=v6.0";
#else
    const string TFM = ".NETCoreApp,Version=v3.1";
#endif

var api = @"... TargetFramework(""${TFM}"", ...";
```
```

stakx (approved):
Looks good to me.

stakx:
> Also NET5 (and even NET 6.0.100) has completely different output from NET6

Like @Shane32 I also suspect this may be related with `[CompilerGenerated]` attribute. .NET 6 emits record types differently than .NET 5 did (e.g. it emits clone methods using covariant returns), that may be one of those differences. There's likely nothing that can be done about the "superfluous" members appearing for .NET 5.

.github/workflows/ci.yml
sungam3r: The reason is to be sure that public api set is the same for all of them.

src/PublicApiGeneratorTests/ApiGeneratorTestsBase.cs
sungam3r: This is only necessary for convenient work with logs. When an error like this occurs then without this output it is almost always hard to find all the necessary parts in GH actions logs.

> Using Console.WriteLine durring unit test execution doesn't seem like the cleanest thing to do

In fact, this is more readable than `Assert.Equal`. It shows only small part of the whole picture and honestly, I often do not understand what is wrong.
изображение

sungam3r:
> There's likely nothing that can be done about the "superfluous" members appearing for .NET 5.

Yes. Let's consider them as a "normal" parts of public api set.

sungam3r:
Sorry for delay.

sungam3r:
@stakx OK to merge?

sungam3r:
@stakx ?

danielmarbach:
@sungam3r I could merge this since I still have contributor rights but I can't release it. Let me know

sungam3r:
As you wish. We have already upgraded our CI to 6.0.402 and 7.0.101.

stakx:
@sungam3r, my apologies for letting this PR sit around for such a long time. I've lately been unexpectedly busy at work and needed some time off from programming-related activities in my spare time. I'm now finally going through all the PRs that have accumulated since.

Sorry for that again!

sungam3r:
> busy at work and needed some time off from programming-related activities in my spare time

The same thing.

sungam3r:
@stakx To keep history clean I suggest to use squash option
изображение

Current history with merge-commits:
изображение

And also this
изображение

stakx:
@sungam3r, I agree that squashing these commits would have resulted in a cleaner commit history, and I was almost going to squash them... but then decided against it at the last minute because I didn't want to risk erasing the commit (co-) authors.

sungam3r:
> I didn't want to risk erasing the commit (co-) authors

This info is preserved. Before merge you can edit final commit message. Usually I just CTRL+A and DELETE all that stuff. As a rule only PR title matters and flows into final commit message... but again - you can change it each time as you want.

stakx:
I knew that I could edit the final commit message, but I wasn't sure about the co-authorships, which cannot be edited through GitHub's webpage. Good to know it's preserved... thanks for letting me know! :+1:

sungam3r:
Nevertheless, it's better to check twice :).

Issue #210, opened 2020-10-05
Looking for new maintainers


danielmarbach:
Originally I stepped in to fill a gap when @JakeGinnivan was looking for people that help to keep the project alive and kicking. That has already been since around 2016, time flies. Now I'm looking for a way to step out of the project gradually if possible. 

Is anyone interested in moving things forward like keep on maintaining the lib, triaging issue, and eventually built-in support for new language features?

danielmarbach:
Pinging @jnm2 @sungam3r @stakx @atifaziz @adamralph to see if anyone of you is interested (since you contributed quite a bit to this library)

stakx:
@danielmarbach, first of all, thank you for having maintained this library&mdash;which we at castleproject/Core have been using for a while&mdash;for the last few years! You were always very quick to respond and put out patch releases for us, which is greatly appreciated! 👌

As to your request, I can see myself lending a hand here. What's holding me back from committing fully at this moment is the fact that this library might get increasingly harder to maintain and keep up-to-date, being based on CodeDOM as it is; the gap between Roslyn's and CodeDOM's capabilities is likely to widen as time goes on. I remember seeing a discussion contemplating a complete rewrite and the back-compat problems that would likely come with it; while this might eventually become inevitable, I'm not sure if I'd be ready to tackle that&mdash;at least not just on my own. I'll give it some more thought, though.

Having said that, I _am_ happy to help out. If you're willing to give me (and others) access to this repository along with some onboarding instructions (well, as far as they are needed), I'll humbly accept. 🙃

sungam3r:
Unfortunately, I also believe that CodeDom is a dead end way. Its use is the main problem in the evolution of the project. It is already clear how much had to be done to bypass its restrictions. Just look at the magic string replacements. It's funny and sad at the same time.

adamralph:
Sorry, I'm afraid I can't take it on.

danielmarbach:
> As to your request, I can see myself lending a hand here. What's holding me back from committing fully at this moment is the fact that this library might get increasingly harder to maintain and keep up-to-date, being based on CodeDOM as it is; the gap between Roslyn's and CodeDOM's capabilities is likely to widen as time goes on. I remember seeing a discussion contemplating a complete rewrite and the back-compat problems that would likely come with it; while this might eventually become inevitable, I'm not sure if I'd be ready to tackle that

[...]

PR #255, opened 2022-12-05
Update CI and project properties

### LINKED ISSUE #254: CI build fails because of `.netcoreapp3.1` TFM
Probably shouldn't have merged #252 and #253 without further checks. Shouldn't be too difficult to fix, but let's do so before merging anything else.

sungam3r:
fixes #254 

Changes:
- dependabot tracks GitHub actions
- CI works entirely on GitHub actions, cmd/sh scripts removed
- SourceLink added
- Description for tool package added
- net7 TFM for tool added
- Tests run for netcoreapp3.1, net6 and net7

~::WARNING:: need to add `NUGET_AUTH_TOKEN` into repository or org secrets to publish packahes to nuget.org~


stakx:
@sungam3r, apologies again for letting this sit around for so long. I haven't reviewed this one yet, but if you'd still like to make some changes to the CI configuration, could you please rebase this on top of current `master`?

sungam3r:
@stakx Done. Let's discuss `Publish Nuget packages to Nuget registry`  step. How did you publish to Nuget.org before? I can remove this step at all. It should not be run on CI (at least preview packages should be published at GitHub package registry). Or I can add other workflow like this [one](https://github.com/graphql-dotnet/graphql-dotnet/blob/master/.github/workflows/publish-code.yml).

stakx:
> How did you publish to Nuget.org before? I can remove this step at all. It should not be run on CI[.]

I have only published a single release so far, and I had to do that manually... which isn't ideal. I completely agree with you that publishing should happen automatically by CI, whenever a version tag is pushed.

sungam3r:
Different strategies exist. I see you use MinVer here. This means automatic version control if I understand correctly. I'm not familiar with MinVer but I know some repos that use it. Personally I prefer semi-automatic (or semi-manual if you like) versioning. Just look at workflows here https://github.com/graphql-dotnet/parser/tree/master/.github/workflows to make an idea. If you like it then I can switch and delete MinVer reference. Otherwise, we need to understand how to integrate MinVer features into CI/publishing nugets.

stakx:
@sungam3r, I haven't studied the CI setup in detail yet. I'll have to give it a closer look. TBH I may not even have a clear preference, as long as the release process it simple. Ideally we could just create a `vX.Y.Z` tag (either by manually pushing it to the repo, or by creating a release on the GitHub webpage) and CI would then release version X.Y.Z for us.

Let me take a closer look at MinVer and the thingy you linked to above. And perhaps in the meantime let me know if you have a specific rexommendation for a release workflow.

stakx:
Alright, I've taken that look.

What I like about MinVer is the way how it takes care about some things like setting `[AssemblyVersion]` etc. I wouldn't want to do that manually.

I also like how you trigger the NuGet publishing process in your `graphql-dotnet/parser` example using `on.release.types: [ 'published' ]`. (If we keep MinVer around, which I think is my current preference, it might make more sense to change the trigger to one that also requires a version tag.)

One thing I think isn't strictly necessary is uploading the `.nupkg` as a GitHub release artifact. They're already on NuGet, and it's much easier to install them from there than downloading a package and putting it in some package source folder. Or am I missing something?

So if we take those things together, why don't we:

* Do releases by creating a release (including a tag in the Git repo) on GitHub;
* which triggers a GitHub Actions pipeline that builds, tests, and pushes the NuGet package;
* and we keep MinVer to put the right version numbers in the assembly and NuGet manifest.

That being said, I'm open to suggestions. You've maybe done this more often than I have.

(Why aren't you a maintainer at this repo btw.? Given your long-standing involvement and interest in this project, that would seem to make sense. :wink:)

sungam3r:
>  They're already on NuGet ... Or am I missing something?

I think no. Someone writes PR. CI triggers, builds code, runs tests. No nuget publishing here, or you will end up with hundreds trash versions of package on nuget.org. OK, then after review and ~20 new commits you merge this PR in `master` - "stable" branch so you **may** have desire to publish package. Where? Either nuget.org or some other artifacts repo (like buffer for work in progres). Really, it is up to you - publish to nuget.org **each** commit to `master` or to do it manually only sometimes accumulating changes from 2/3/11/45 commits. In graphql-dotnet org repos we chose the second strategy for publishing to nuget.org but **each** commit to `master` or `develop` still results in publishing packages to nuget.pkg.github.com . Why? Because someone may have a strong need to obtain code-fixes (or test features) from these "internal" releases before "official external" release appears on GitHub/nuget.org. It is possible that for the first strategy MinVer will be a suitable decision - I do not know.

>  Do releases by creating a release (including a tag in the Git repo) on GitHub;
    which triggers a GitHub Actions pipeline that builds, tests, and pushes the NuGet package;
    and we keep MinVer to put the right version numbers in the assembly and NuGet manifest.

Well, then I suggest to remove `Publish Nuget packages to Nuget registry` step from ci workflow for now and merge as is (after your review) since proposed CI should be equal to all scripts before. You can publish releases the same way as you did before - manually. Further activity about publishing process (CD) will be tracked here - #258.


> Why aren't you a maintainer at this repo btw

No one invited me.

stakx:
> No one invited me.

I see. Perhaps it wasn't clear in #210 that you were interested, too? If you still are, let me know.

>Someone writes PR. CI triggers, builds code, runs tests. No nuget publishing here, or you will end up with hundreds trash versions of package on nuget.org.

Oh, I think there's a misunderstanding. Of course publishing every single commit on `master` to NuGet would be overkill. I was only referring to a CI pipeline that publishes to NuGet; that one wouldn't have to publish the same NuGet package as a GitHub release artifact.

> we chose the second strategy for publishing to nuget.org but each commit to `master` or `develop` still results in publishing packages to `nuget.pkg.github.com`. Why? Because someone may have a strong need to obtain code-fixes (or test features) [...]

I see what you mean, makes sense. Not sure if it's really required for PublicApiGenerator which, I believe, has had relatively quick release cycles / iterations so users probably don't have to wait for very long for an official release (and they can always just as for one). It's not like we're planning several milestones far into the future, we're mostly doing small bugfixes or improvements that can be published more or less right away.

That being said, I have no problem with the release scheme you described.

It's getting late here, I'll need to postpone the review to tomorrow or the day after.

sungam3r:
>  If you still are, let me know.

@stakx I could help to move things a bit faster here.

src/Directory.Build.props
```diff
@@ -2,26 +2,15 @@
...
-    <PackageOutputPath>..\..\nugets</PackageOutputPath>
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
-  </PropertyGroup>
-
-  <PropertyGroup>
     <MinVerMinimumMajorMinor>10.0</MinVerMinimumMajorMinor>
-    <NoWarn>NU5105</NoWarn>
+    <NoWarn>$(NoWarn);NU5105</NoWarn>
+    <IsPackable>true</IsPackable>
```
stakx, line 11: Making projects "packable" by default seems a little dangerous. Say that in the NuGet publishing step, we simply scan for all `*.nupkg` and publish them, and say that for some reason we add a second unit test project in the future but we forget to explicitly set its `<IsPackable>` property to `False`; we'd then publish the unit tests accidentally, wouldn't we?

I believe it would be safer to explicitly opt-in to `<IsPackable>`, rather than having to opt-out.

stakx (commented):
Overall, these changes look plausible and sensible to me.

I haven't yet had the time to actually run a build with these changes, but am I understanding correctly that...

* You're removing `build.{sh,cmd}` because your MSBuild refactor means one can simply run `dotnet build`  directly instead?

* You're moving certain repeated project settings (such as the setup for strong-naming assemblies) into `Directory.Build.{props,targets}`?

sungam3r:
> You're removing build.{sh,cmd} because your MSBuild refactor means one can simply run dotnet build directly instead?

Yes and no. These files IMO are needed only for CI on the remote machine. I "unwrapped" their contents into workflow. Of course developer **may** run such files locally as well but the need for this is extremely unlikely.

> You're moving certain repeated project settings (such as the setup for strong-naming assemblies) into Directory.Build.{props,targets}?

Yes. I am used to following this strategy in all the projects which I work on.

src/Directory.Build.props
sungam3r: In fact, I spent a lot of time to comprehend what strategy for this property (in conjunction with others) is the most correct and universal for any .NET repo. There are several variations that one way or another lead to the desired result. For such a small project as this I can apply another option that you are talking about.

sungam3r: done

stakx:
I may still be missing something, but now that you've removed `<IsPackable>` everywhere (along with `<GeneratePaclageOnBuild>`) will `.nupkg`s still be produced where appropriate? Also it seems that since you've made strong-naming conditional (i.e. dependent on `<IsPackable>`) you essentially also stopped strong-naming the assemblies now...?

sungam3r:
> I may still be missing something, but now that you've removed <IsPackable> everywhere (along with <GeneratePackageOnBuild>) will .nupkgs still be produced where appropriate?

Yes, because `IsPackable` by default is true for libraries and global tools. Also `IsPackable` is false for all test projects- i.e. projects that reference `Microsoft.NET.Test.Sdk` package. MS took care of a reasonable system of default values.

> Also it seems that since you've made strong-naming conditional (i.e. dependent on <IsPackable>) you essentially also stopped strong-naming the assemblies now...?

All test projects are not signed (as well as samples, dummy assemblies if any in the future). All other projects (packable ones) are signed.

sungam3r:
@stakx OK to merge?

src/PublicApiGenerator.Tool/PublicApiGenerator.Tool.csproj
```diff
@@ -1,14 +1,11 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp2.1;net7</TargetFrameworks>
```
stakx, line 5: Just wondering whether or not we could pick better targets here. `netcoreapp2.1` is relatively ancient, do we still need that? (I don't mind but would probably have picked `netcoreapp3.1` instead.) And instead of `net7` I'd probably choose `net6.0` which is an LTS release, and should work for .NET 7, too.

stakx:
@sungam3r, I suppose so. I'll leave this in your capable hands. If we notice any CI issues, we can always improve those incrementally.

stakx (approved):

stakx:
Feel free to go ahead and merge.

Just one unrelated request: I notice that you disabled merge commits and rebases. Could we please undo this? I disagree with squash-merging only; the type of merge should be a case-by-case decision.

src/PublicApiGenerator.Tool/PublicApiGenerator.Tool.csproj
sungam3r: I'm fine either way. I changed to 3.1 + 6.

sungam3r:
I did it so to keep history clean. If you look at https://github.com/PublicApiGenerator/PublicApiGenerator/commits/master then you will see commits with messages like `Merge pull request #xxx from yyy` or even worse:
изображение

These are temporary commits, which should not be in the master branch, they only clog history and do not provide a holistic picture of the work done, breaking it into many small pieces.

> the type of merge should be a case-by-case decision.

😕 I don't know. In practice, I have not seen cases when I had to choose between merge/rebase and squash.I returned merge/rebase options back but I will never use them for my own PRs.

sungam3r:
@stakx Regarding TFMs:
изображение

IMO that was a reason why I chose 2.1 over 3.1.

PR #257, opened 2022-12-30
Add support for C# 9 init-only setters

### LINKED ISSUE #250: Init-only setters should be marked as `init;`
As it is not permissible for code compiled with C# 8 or prior to access init-only setters, the generated public API text should discriminate normal setters (`set;`) from init-only setters (`init;`).

stakx:
Closes #250.

Not sure if this is the best way to get this done (probably not!), but for now it appears to work.


src/PublicApiGenerator/CodeNormalizer.cs
```diff
@@ -24,6 +24,7 @@ internal static class CodeNormalizer
         internal const string EventRemovePublicMarker = "removepublic";
         internal const string MethodModifierMarkerTemplate = "_{0}_3C0D97CD952D40AA8B6E1ECB98FFC79F_";
         internal const string PropertyModifierMarkerTemplate = "_{0}_5DB9F56043FF464997155541DA321AD4_";
+        internal const string PropertyInitOnlySetterTemplate = "_{0}_156783F107B3427090B5486DC33EE6A9_";
```
danielmarbach, line 27: Hahah, all my hacks are still alive

PR #278, opened 2023-02-26
Fix CodeQL issues

sungam3r:


src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -565,8 +562,9 @@ private static CodeExpression CreateInitialiserExpression(CustomAttributeArgumen
         {
             // CodeDOM outputs a verbatim string. Any string with \n is treated as such, so normalize
             // it to make it easier for comparisons
-            value = Regex.Replace(s, @"\n", "\\n");
```
sungam3r, line 568: From CodeQL: This assignment to is useless, since its value is never read.


so bugfix here

PR #281, opened 2023-02-27
Fix readme in nupkg

### LINKED ISSUE #277: For some reason nuget.org says that package does not contain readme
изображение

sungam3r:
fixes #277 

It was due to how/when MSBuild resolves `IsPackable` value.


src/Directory.Build.props
```diff
@@ -14,6 +14,7 @@
     <EnableNETAnalyzers>true</EnableNETAnalyzers>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <WarningsNotAsErrors>$(TreatWarningsNotAsErrors);1591</WarningsNotAsErrors>
```
sungam3r, line 17: temporary

PR #286, opened 2023-02-28
Introduce two new options

### LINKED ISSUE #184: Question about WhitelistedNamespacePrefixes
Is it conceived that this property does not affect the printing of the class itself? I observe empty classes in the output. 

`WhitelistedNamespacePrefixes = new[] { }`:
```c#
namespace System.IdentityModel.Tokens.Jwt
{
    public static class JwtSecurityTokenExtensions { }
}
```

 `WhitelistedNamespacePrefixes = new[] { "System.IdentityModel.Tokens.Jwt" }`:
```c#
namespace System.IdentityModel.Tokens.Jwt
{
    public static class JwtSecurityTokenExtensions
    {
        public static bool CheckGrantOperation(this System.IdentityModel.Tokens.Jwt.JwtSecurityToken token, string operation) { }
        public static string[] GetGrantOperations(this System.IdentityModel.Tokens.Jwt.JwtSecurityToken token) { }
        public static string GetUserLogin(this System.IdentityModel.Tokens.Jwt.JwtSecurityToken token) { }
    }
}
```

sungam3r:
Fixes #184 in two aspects:
1. It never prints "empty" classes like that
```c#
namespace System.IdentityModel.Tokens.Jwt
{
    public static class JwtSecurityTokenExtensions { }
}
```
Those classes are completely excluded from output, i.e. whitelisting works now for types as well.

2. If one wants to output extension methods even from "blacklisted" namespaces then he may set `options.UseBlacklistedNamespacePrefixesForExtensionMethods = false` 

Also I moved hardcoded `System` and `Microsoft` prefixes into public `BlacklistedNamespacePrefixes` option.

@danielmarbach @stakx please review.

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -1,11 +1,11 @@
-using Microsoft.CSharp;
-using Mono.Cecil;
-using Mono.Cecil.Rocks;
 using System.CodeDom;
 using System.CodeDom.Compiler;
 using System.Globalization;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using Microsoft.CSharp;
+using Mono.Cecil;
+using Mono.Cecil.Rocks;
 using ICustomAttributeProvider = Mono.Cecil.ICustomAttributeProvider;
 using TypeAttributes = System.Reflection.TypeAttributes;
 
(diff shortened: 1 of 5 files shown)


src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -1,11 +1,11 @@
-using Microsoft.CSharp;
```
sungam3r, line 1: It's worth to use _hide whitespaces_ option when reviewing.

src/PublicApiGeneratorTests/Namespaces_whitelisting.cs
```diff
@@ -32,12 +32,7 @@ public class Simple2
         [Fact]
         public void Should_filter_microsoft_namespace()
         {
-            AssertPublicApi(new[] { typeof(Simple1), typeof(Simple2) },
```
sungam3r, line 35: I have to change 2 tests here to align with new behavior. I do not think that it should be treated as a breaking change. It's more like a bugfix since while/black-listing should work for entire classes as well.

danielmarbach:
I wonder if it would be a good time to replace "Whitelist" with "Allowlist" and "Blacklist" with "Denylist" or something similar.

sungam3r:
@danielmarbach I would like to merge this ASAP and move on publishing release. Ping me back with your comments if any.

danielmarbach:
I would rename the current APIs to make it more inclusive but I'm no longer a maintainer nor do I want to impose my views on you. Feel free to ignore me

PR #288, opened 2023-02-28
Escape literals in public constant string

### LINKED ISSUE #82: Escape literals in public constant string
Currently, API generator escapes `\n` and `\r\n` in attribute strings but it doesn't escape literals in public constant fields like

```
public const string MyString = @"This
Is
Multiline
";
```

sungam3r:
fixes #82


sungam3r:
ping @danielmarbach

danielmarbach (approved):

PR #342, opened 2023-12-03
Fix hiding of member with different return type

sungam3r:
Transferred from https://github.com/PublicApiGenerator/PublicApiGenerator/pull/226

изображение


sungam3r:
@stakx This one is ready for review.

изображение

4 years....

PR #367, opened 2024-02-15
Update the Verify.Xunit example

peymanr34:
The `UsesVerify` attribute is [no longer required](https://github.com/VerifyTests/Verify/pull/1131) and it can be removed.


sungam3r:
Thanks.

PR #391, opened 2024-09-13
Order types by Namespace before FullName

### LINKED ISSUE #389: Unexpected ordering on FullName
When dumping this API, the namespace `A.C` is listed before `A` as we compare the types by `FullName` and `"A.C.B"` comes before `"A.D"`.
```cs
namespace A
{
    public class D;
}

namespace A.C
{
    public class B;
}
```

```
[assembly: System.Runtime.Versioning.TargetFramework(".NETCoreApp,Version=v8.0", FrameworkDisplayName=".NET 8.0")]
namespace A.C
{
    public class B
    {
        public B() { }
    }
}
namespace A
{
    public class D
    {
        public D() { }
    }
}
```

If I then add a new type `A.A`, the dump now lists the namespace `A` before `A.C` as `"A.A"` comes before `"A.C.B"`.

```cs
namespace A
{
    public class A;
}

namespace A
{
    public class D;
}

namespace A.C
{
    public class B;
}
```

```
[assembly: System.Runtime.Versioning.TargetFramework(".NETCoreApp,Version=v8.0", FrameworkDisplayName=".NET 8.0")]
namespace A
{
    public class A
    {
        public A() { }
    }
    public class D
    {
        public D() { }
    }
}
namespace A.C
{
    public class B
    {
        public B() { }
    }
}
```

The real issue for me is not _how_ the types are ordered but that adding a single type to one namespacce will re-order _other_ namespaces, which makes it harder to diff.
As I understand from reading #56 and #138 one goal for this project is to make the output friendly for diffing.
Since GitHub uses the myers diff algorithm, which is not great with moving elements, the diff can be quite confusing.

A suggestion is to create a custom `TypeReference` comparer and first compare by `Namespace` and then by `Name`.
I'm trying that out [here](https://github.com/jnyrup/PublicApiGenerator/tree/jn/type_ordering)
This works for the case above, but I note that `TypeReference.FullName` also looks at `IsNested` and whether `NameSpace` is null or empty - I haven't thought about how this should be compared.

Here's a test to reproduce:
```cs
[Fact]
public void Should_order_namespaces_alphabetically2()
{
    AssertPublicApi(
    [
        typeof(A.D),
        typeof(A.C.B),
    ],
@"namespace PublicApiGeneratorTests.A
{
    public class D
    {
        public D() { }
    }
}
namespace PublicApiGeneratorTests.A.C
{
    public class B
    {
        public B() { }
    }
}");
}

namespace A
{
    public class D;
}

namespace A.C
{
    public class B;
}
```

jnyrup:
This fixes #389

### DIFF
--- a/src/PublicApiGenerator/ApiGenerator.cs
+++ b/src/PublicApiGenerator/ApiGenerator.cs
@@ -91,12 +91,12 @@ public static class ApiGenerator
         var compileUnit = new CodeCompileUnit();
         if (options.IncludeAssemblyAttributes && assembly.HasCustomAttributes)
         {
-            PopulateCustomAttributes(assembly, compileUnit.AssemblyCustomAttributes, attributeFilter);
+            PopulateCustomAttributes(assembly, compileUnit.AssemblyCustomAttributes, attributeFilter, options.TypeComparer);
         }
 
         var publicTypes = assembly.Modules.SelectMany(m => m.GetTypes())
             .Where(shouldIncludeType)
-            .OrderBy(t => t, new TypeReferenceComparer());
+            .OrderBy(t => t, options.TypeComparer);
         foreach (var publicType in publicTypes)
         {
             var @namespace = compileUnit.Namespaces.Cast<CodeNamespace>().FirstOrDefault(n => n.Name == publicType.Namespace);
@@ -108,7 +108,7 @@ public static class ApiGenerator
 
             using (NullableContext.Push(publicType))
             {
-                var typeDeclaration = CreateTypeDeclaration(publicType, options.DenyNamespacePrefixes, options.AllowNamespacePrefixes, options.UseDenyNamespacePrefixesForExtensionMethods, attributeFilter);
+                var typeDeclaration = CreateTypeDeclaration(publicType, options.DenyNamespacePrefixes, options.AllowNamespacePrefixes, options.UseDenyNamespacePrefixesForExtensionMethods, attributeFilter, options.TypeComparer);
                 @namespace.Types.Add(typeDeclaration);
             }
         }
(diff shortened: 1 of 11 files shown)


sungam3r:
@jnyrup Thanks for contribution. I'm sure it would be better to opt-in in such feature since for many users it will be breaking change forcing them to reorder approval files. Please add new option in `ApiGeneratorOptions` like `public OrderMode OrderBy { get; set; } = OrderMode.FullName;`

```
public enum OrderMode
{
    FullName,
    NamespaceThenFullName,
}
```

jnyrup:
Thanks for the feedback.
Opt-in does indeed seems like a good compromise 👍

I've went ahead and implemented the suggestion with an `OrderMode` enum exposed in `ApiGeneratorOptions`.

What do you think about instead generalizing this to let the user provide the `TypeReference` comparer?
The downside being that we then reference `Mono.Cecil.TypeReference` in the PublicApiGenerator API.

```cs
public Comparer<TypeReference> TypeComparer { get; set; } = new FullNameComparer();
```

Where `FullNameComparer` would be 
```cs
internal sealed class FullNameComparer : Comparer<TypeReference>
{
    public override int Compare(TypeReference x, TypeReference y) =>
        StringComparer.Ordinal.Compare(x.FullName, y.FullName);
}
```

sungam3r:
> What do you think about instead generalizing this to let the user provide the TypeReference comparer?

I would not. All Mono.Cecil stuff is internal implementation detail.

src/PublicApiGenerator/OrderMode.cs
```diff
@@ -0,0 +1,16 @@
...
+/// <summary>
+/// Used to instruct the generator how to order types.
+/// </summary>
+public enum OrderMode
+{
+    /// <summary>
+    /// Order types by <see cref="Type.FullName"/>
+    /// </summary>
+    FullName = 0,
+    /// <summary>
```
sungam3r, line 12:
```suggestion

    /// <summary>
```

src/PublicApiGenerator/ApiGeneratorOptions.cs
```diff
@@ -81,4 +83,23 @@ public class ApiGeneratorOptions
...
+    {
+        get => _orderBy;
+        set
+        {
+            _orderBy = value;
+            TypeComparer = value == OrderMode.NamespaceThenFullName
+                ? new NamespaceThenFullNameComparer()
+                : new FullNameComparer();
+        }
+    }
```
sungam3r, line 102:
```suggestion
    /// <summary>
    /// Instructs the generator how to order types. Defaults to <see cref="OrderMode.FullName"/>
    /// </summary>
    public OrderMode OrderBy
    {
        get => TypeComparer is FullNameComparer ? OrderMode.FullName : OrderMode.NamespaceThenFullName;
        set
        {
            TypeComparer = value == OrderMode.NamespaceThenFullName
                ? new NamespaceThenFullNameComparer()
                : new FullNameComparer();
        }
    }
```

src/PublicApiGenerator/ApiGeneratorOptions.cs
```diff
@@ -81,4 +83,20 @@ public class ApiGeneratorOptions
...
+    /// Instructs the generator how to order types. Defaults to <see cref="OrderMode.FullName"/>
+    /// </summary>
+    public OrderMode OrderBy
+    {
+        get => TypeComparer is FullNameComparer ? OrderMode.FullName : OrderMode.NamespaceThenFullName;
+        set
+        {
+            TypeComparer = value == OrderMode.NamespaceThenFullName
+                ? new NamespaceThenFullNameComparer()
+                : new FullNameComparer();
```
sungam3r, line 97:
```suggestion
            TypeComparer = value == OrderMode.FullName
                ? new FullNameComparer()
                : new NamespaceThenFullNameComparer();
```

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -290,7 +291,7 @@ private static CodeTypeDeclaration CreateTypeDeclaration(TypeDefinition publicTy
                 declaration.BaseTypes.Add(publicType.BaseType.CreateCodeTypeReference(publicType));
             }
         }
-        foreach (var @interface in publicType.Interfaces.OrderBy(i => i.InterfaceType.FullName, StringComparer.Ordinal)
+        foreach (var @interface in publicType.Interfaces.OrderBy(i => i.InterfaceType, typeComparer)
```
sungam3r, line 294: I'm not sure about it. This sorts implemented interfaces like `A : I1, I2`. I think sorting should always work as `.OrderBy(i => i.InterfaceType.FullName, StringComparer.Ordinal)`.

sungam3r: The same for generic parameters and type members. In other words new setting should work only for "top-level entities" - types/namespaces.

jnyrup:
Addressed the comment about only sorting top-level types and Code Coverage should also be better now - let's see what CI thinks.

sungam3r:
Append new enum in `ExcludeTypes` list inside `Should_exclude_types` test.

jnyrup:
> Append new enum in `ExcludeTypes` list inside `Should_exclude_types` test.

Thanks, had branched of before #394

sungam3r:
https://github.com/PublicApiGenerator/PublicApiGenerator/releases/tag/11.3.0

PR #393, opened 2024-11-05
Update README.md

peymanr34:
Added a comment about multiple target frameworks for the `Verify.Xunit` example.


sungam3r:
Thanks.

PR #394, opened 2024-12-08
Allow to explicitly exclude types

### LINKED ISSUE #388: Allow specific types to be excluded
Hi

I've just started to use this library but have encountered an unfortunate issue.

In our API, we expose a public type that is shared between our consumers. However, it is not technically public in the same sense as methods and other types in our API. Our consumers use this type to communicate the global availability of features used for feature flagging, and the type resides in our API to allow easier distribution.

As I can understand from this issue #24, it was decided that such functionality should not be added because any change to the public API is considered a breaking change.
While that makes sense, it's more complex in my specific case.

Right now, I'm using a workaround where I manually remove the type using regex. Still, I'd be more comfortable simply excluding the type directly through the options, similar to how attributes can be excluded (#45).

Are there any plans to extend the options to allow excluding specific types?

From the implementation, adding support for this would be relatively straightforward, and I would also be willing to provide a pull request if the request sounds reasonable.

### LINKED ISSUE #395: Add "XamlGeneratedNamespace" to list of default excluded namespaces
When developing apps using WPF and in debug mode, the tooling (sometimes) adds a `XamlGeneratedNamespace` namespace, "breaking" our tests. It does not generate this namespace in release mode.

I think this is only used during debug for tooling and is not part of the public API. Would you be open to a PR where this namespace is added to the list of default namespaces?

sungam3r:
fixes #388
fixes #395


src/PublicApiGeneratorTests/PublicApiGeneratorTests.csproj
```diff
@@ -26,7 +26,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
     <PackageReference Include="xunit" Version="2.8.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.8.0" NoWarn="NU1701">
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.5.7" NoWarn="NU1701">
```
sungam3r, line 29: I had to revert it since VS test discoverer was broken.

sungam3r:
@siewers please review that this is what you want

sungam3r:
@stakx @danielmarbach Any objections?

siewers:
@sungam3r, I'm sorry for not getting back to you sooner. It looks exactly like what I need. Thank you!

PR #411, opened 2025-01-03
Add failing test for long string attribute arguments

siewers:
Added a test that shows how long strings as attribute arguments will be surrounded by parenthesis.


sungam3r:
@siewers In #412 I can not reproduce it.

siewers:
That's strange. Let me check if I made a mistake. It was failing locally on my side.

sungam3r:
My bad. Thanks. Replaces by #412.

PR #412, opened 2025-01-03
Fix doubled braces in attributes

### LINKED ISSUE #410: Version 11.3.0 `ObsoleteAttribute` behavior differs from 11.2.0
I've just tried upgrading to version 11.3.0 and see changes in the `ObsoleteAttibute` behavior, where the received version has additional parenthesis around the message string.

Version 11.2.0:
```cs
[System.Obsolete("Constructors of types with required members are not supported in this version of your compiler.", true)]
```

Version 11.3.0:
```cs
[System.Obsolete(("Constructors of types with required members are not supported in this version of your compiler."), true)]
```

Do you know if this is expected?

There might be other cases, but these were the only ones I've seen.

sungam3r:
fixes #410
_____

This PR was superseded by #414.


sungam3r:
Wow! Not reproducable locally, but reproducable in CI.

sungam3r:
There are no doubled (( on NET FRAMEWORK.

siewers:
Okay, then it makes more sense.
I'm not using .NET Framework so I did not test that 🙂

src/PublicApiGenerator/ApiGenerator.cs
```diff
@@ -496,7 +496,7 @@ private static string ConvertAttributeToCode(Func<CodeTypeReference, CodeTypeRef
         var attribute = GenerateCodeAttributeDeclaration(codeTypeModifier, customAttribute);
         var declaration = new CodeTypeDeclaration("DummyClass")
         {
-            CustomAttributes = new CodeAttributeDeclarationCollection(new[] { attribute }),
+            CustomAttributes = new CodeAttributeDeclarationCollection([attribute]),
```
sungam3r, line 499:
```suggestion
            CustomAttributes = new CodeAttributeDeclarationCollection(new[] { attribute }),
```

sungam3r:
@stakx @jnyrup please take a look. I'll keep it 1-2 days before merging. I woul like to release this with one with #342.

src/PublicApiGenerator/CodeNormalizer.cs
```diff
@@ -32,6 +37,13 @@ public static string NormalizeGeneratedCode(StringWriter writer)
     {
         var gennedClass = writer.ToString();
 
+        // https://github.com/PublicApiGenerator/PublicApiGenerator/issues/410
+        if (gennedClass.Contains(ATTRIBUTE_MARKER_AND_DOUBLED_BRACE) && gennedClass.Contains(DOUBLED_CLOSING_BRACE_IN_ATTRIBUTE))
+        {
+            gennedClass = gennedClass.Replace(ATTRIBUTE_MARKER_AND_DOUBLED_BRACE, ATTRIBUTE_MARKER_AND_BRACE);
+            gennedClass = gennedClass.Replace(DOUBLED_CLOSING_BRACE_IN_ATTRIBUTE, SINGLE_CLOSING_BRACE_IN_ATTRIBUTE);
+        }
```
jnyrup, line 45: This doesn't work with the example from #410 because it has an extra parameter after the long string.

```cs
        [Fact]
        public void Should_handle_attribute_with_long_string_initialiser2()
        {
            AssertPublicApi<ClassWithAttributeWithLongStringInitialiser2>(
                @"namespace PublicApiGeneratorTests.Examples
{
    [System.Obsolete(""This is a veeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"" +
        ""eeeeeeery long string that will be surrounded by parenthesis due to how it\'s han"" +
        ""dled by Microsoft.CodeDom"", false)]
    public class ClassWithAttributeWithLongStringInitialiser2
    {
        public ClassWithAttributeWithLongStringInitialiser2() { }
    }
}");
        }
```

```cs
[Obsolete("This is a veeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeery long string that will be surrounded by parenthesis due to how it's handled by Microsoft.CodeDom", false)]
public class ClassWithAttributeWithLongStringInitialiser2()
{
}
```

There's also other cases to consider:
* Ending with long string parameter `[SomeAttribute(false, "loooong string")]`
* Multiple long strings parameters `[SomeAttribute("loooong string", "loooong string")]`


Looking at the [commit history](https://github.com/dotnet/runtime/commits/main/src/libraries/System.CodeDom) for `System.CodeDom` https://github.com/dotnet/runtime/pull/96948 is the only change between 8.0.0 and 9.0.0.
Skimming the fixed issue, I'm not sure if the fix is relevant for PublicApiGenerator.

So how about:
* Downgrade `System.CodeDom` to 8.0.0
* Undo the changes to `CodeNormalizer`
* Keep `Should_handle_attribute_with_long_string_initialiser` as a regression test

sungam3r: You are right. It turned out to be not so easy to fix this error only by replacing lines. As a result, I copied all needed code to generate results and commented out problem line.

sungam3r: All tests are green now.

> Downgrade System.CodeDom to 8.0.0

I understand but downgrading **may** lead to additional noise with dependabot.

jnyrup: I think dependabot will keep silent if we fix the version using `Version=[8.0.0]`.

sungam3r: I know that feature. Nevertheless I'm 50/50 now to go on with these changes or revert to 8.

sungam3r: @jnyrup I decided to revert System.CodeDom to 8, publish release and keep #412 opened just in case.

sungam3r:
@jnyrup I achieved a minimum possible diff to fix this issue.

PR #428, opened 2025-01-07
Handle init-only properties gracefully

sungam3r:


src/PublicApiGenerator/System.CodeDom/CSharpCodeGenerator.cs
```diff
@@ -2930,7 +2948,7 @@ public string GetTypeOutput(CodeTypeReference typeRef)
 
         private void OutputStartingBrace()
         {
-            if (_options.BracingStyle == "C")
+            if (_options.BracingStyle == "Block")
```
sungam3r, line 2951: Was inverted ? 😕

PR #434, opened 2025-01-08
Support property method modifiers

sungam3r:


src/PublicApiGeneratorTests/Property_visibility.cs
```diff
@@ -170,21 +169,6 @@ public ClassWithPublicGetterPrivateProtectedSetter() { }
 }");
         }
 
-        [Fact(Skip = "Not supported by CodeDOM")]
-        [Trait("TODO", "Property method modifiers not supported by CodeDOM")]
-        public void Should_output_protected_internal_setter_for_public_property_with_correct_modifier()
```
sungam3r, line 175: This test is a duplicate of `Should_output_protected_internal_setter_for_public_property`.

Issue #104, opened 2019-11-01
Using C#-like syntax for `dynamic` (and other keywords that compile to attributes)


jnm2:
Would you be interested in showing `[System.Runtime.CompilerServices.DynamicAttribute()] object` as `dynamic`, similar to the C# syntax that you'd see in intellisense when interacting with the API?

An argument against it might be that it's better to put lower-level realities in your face in case the API is used from a language where the attribute is not specially-recognized.

Maybe an argument for it would be that it would be consistent with `[ParamsAttribute]`, `[ExtensionAttribute]`, etc, and the same kind of thing with C# 8 attributes: https://github.com/ApiApprover/ApiApprover/issues/54

A few other attributes have cropped up like the ones for the C# 7.0 `ref` and `readonly` keywords when declaring a struct:

```cs
public readonly ref struct Foo
{
}
```

Desugared:

```cs
[IsByRefLike]
[Obsolete("Types with embedded references are not supported in this version of your compiler.", true)]
[IsReadOnly]
public struct Foo
{
}
```
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABAzAAlgQwBMIA7AGwE88YAzHAZwBcoBXMRnAMQggFgAoAN4CAvkA=

danielmarbach:
Dynamic sure but I say we can target that for another release. It seems of lower importance than the other cases you mentioned Rudimentary readonly support for example for structs is in https://github.com/ApiApprover/ApiApprover/pull/84But sure the support is probably not yet superb and could be made better

jnm2:
Another one is `volatile` instead of `modreq(System.Runtime.CompilerServices.IsVolatile)`.

sungam3r:
I have already come across this notation (modreq) in #54.

sungam3r:
I think that `volatile` itself is not a part of api. It is just implementation detail.

sungam3r:
Regarding dynamic: https://sharplab.io/#gist:e22ba1bb9e308621e59bce8eaf95c021

The rabbit hole is quite deep...

jnm2:
> I think that `volatile` itself is not a part of api. It is just implementation detail.

The modreq indicates how important it is as API. Modreqs are how you can express "this object requires special handling." 

[...]

PR #436, opened 2025-01-08
Add support for readonly ref struct

sungam3r:


sungam3r:
@jnm2 please review, you posted about it in #104.

